PR #45, opened 2023-03-09
Fixed infinite recursion crash

SamWindell:
Came across this problem when compiling on linux. The templated log() would call itself rather than the regular log().

### DIFF
--- a/src/detail/os/osutil.h
+++ b/src/detail/os/osutil.h
@@ -26,14 +26,15 @@ namespace os
 	std::string getParentFolderName();
 	std::string getBinaryName();
 
+	void log(const char* text);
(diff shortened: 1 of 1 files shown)


defiantnerd:
I've tried to fix this by hinting the type. Does this solve the problem, too?

SamWindell:
I'll give it a try when I'm next using this

SamWindell:
Yes that also fixes it, but I've discovered that the type hinting is not necessary. The important change in your commit is that the non-template `log` is now declared before the template. Thanks!

defiantnerd:
Thanks for bringing this to my attention.

PR #57, opened 2023-08-26
Convert VST3 noteOnEvent.tuning to a CLAP Note Expression

baconpaul:
If the VST3 noteOnEvent.tuning field exists, and if the target wrapped clap supports note expressions, convert the tuning to the appropriate note expression. Among other things, this allows dorico to work with a wrapped surge xt clap in microtonal mode.


defiantnerd:
I like the niftiness of this...

Issue #22, opened 2022-11-01
[vst3] only add IMidiMapping if CLAP supports DIALECT_MIDI


defiantnerd:
Actual MIDI controllers can only be sent to a CLAP if it supports DIALECT_MIDI.
Currently, `IMidiMapping` is always provided, but shouldn't if the CLAP can't receive it.

PR #61, opened 2023-08-27
issue #22: provide IMidiMapping interface dynamically

defiantnerd:
IMidiInterface only when CLAP supports CLAP_NOTE_DIALECT_MIDI.

### DIFF
--- a/src/wrapasvst3.h
+++ b/src/wrapasvst3.h
@@ -145,7 +145,14 @@ public:
 	//---Interface--------------------------------------------------------------------------
 	OBJ_METHODS(ClapAsVst3, SingleComponentEffect)
 	DEFINE_INTERFACES
-	DEF_INTERFACE(IMidiMapping)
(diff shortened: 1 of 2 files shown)


src/wrapasvst3.h
```diff
@@ -222,6 +231,7 @@ class ClapAsVst3 : public Steinberg::Vst::SingleComponentEffect
 	util::fixedqueue<queueEvent, 8192> _queueToUI;
 
 	// for IMidiMapping
+  bool _useIMidiMapping = false;
```
baconpaul, line 234: maybe you want the default here to be true to keep it compatible with the pre-behavior. Minor point.

baconpaul:
Looks great to me!

src/wrapasvst3.h
defiantnerd: If the plugin has no note port or does not allow MIDI, there is no point in IMidiMapping, I guess.

PR #62, opened 2023-08-28
fix: release `bundleURL` after last use of `_bundle`

0ax1:
`bundleURL` is referenced, not copied, by `_bundle` which is why it shouldn't be released before the last use of `_bundle`.


0ax1:
After discussion @baconpaul  we concluded that `_bundle` keeps an inner copy of `bundleURL` which makes this PR obsolete.

PR #63, opened 2023-08-28
Rework the CMake setup

baconpaul:
This set of CMake changes updates the build system to

1. Not use the steinberg cmake library
2. Fix several problems from using the above, include symbol hiding properly on macos and linux, not requiring the xcode generator, making the lin/win bundle code more explicit
3. Turn on linux in the pipeline / ci check also.

I didn't squash these yet since I'm sure ther ewill be review comments and I'll add more.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,7 @@ target_compile_options(${pluginname_vst3} PRIVATE
 		)
 
 if (NOT "${CLAP_VST3_TUID_STRING}" STREQUAL "")
-	message(STATUS "Using cmake-specified VST3 TUID ${CLAP_VST3_TUID_STRING}")
+	message(STATUS "clap-wrapper: Using cmake-specified VST3 TUID ${CLAP_VST3_TUID_STRING}")
 	target_compile_options(${pluginname_vst3} PRIVATE
 			-DCLAP_VST3_TUID_STRING="${CLAP_VST3_TUID_STRING}"
 			)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,15 +2,25 @@
 #	clap-wrapper project
 #
 #   Copyright (c) 2022 Timo Kaluza (defiantnerd)
-#					   Paul Walker
+#					   Paul Walker (baconpaul)
 #
-#	
-
+# There are a variety of advanced options to configure and build this wrapper, but
+# for the simple case of single plugin in a single clap, here are some relevant
+# cmake-time options which this wrapper will use. See the documentation for a more
+# complete description
+#
+# CLAP_WRAPPER_OUTPUT_NAME   The name of the resulting .vst3 or .component
+# CLAP_SUPPORTS_ALL_NOTE_EXPRESSIONS  The wrapper will forward all VST3 note expressions to your CLAP
+# CLAP_VST3_TUID_STRING  The VST3 component ::iid as a string; absent this wrapper hashes clap id
+# CLAP_WRAPPER_BUNDLE_IDENTIFIER the macOS Bundle Identifier. Absent this it is 'org.cleveraudio.wrapper.(name)'
+# CLAP_WRAPPER_BUNDLE_VERSION the macOS Bundle Version. Defaults to 1.0
+#
 
 cmake_minimum_required(VERSION 3.20)
 cmake_policy(SET CMP0091 NEW)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum macOS version")
 set(CMAKE_OSX_ARCHITECTURES "x86_64;arm64" CACHE STRING "Build Universal Always")
+set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>") # Statically link the MSVC Runtime
 
 # If your clap supports note expressions you *can* implement the wrapper extension here or you
 # can just build with this turned on and it will forward all note expressions to your CLAP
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -22,110 +35,106 @@ project(clap-wrapper
...
+	set_target_properties(${pluginname_vst3} PROPERTIES
+			BUNDLE True
+			BUNDLE_EXTENSION vst3
+			MACOSX_BUNDLE_GUI_IDENTIFIER ${CLAP_WRAPPER_BUNDLE_IDENTIFIER}
+			MACOSX_BUNDLE_BUNDLE_NAME ${clapname}
+			MACOSX_BUNDLE_BUNDLE_VERSION ${CLAP_WRAPPER_BUNDLE_VERSION}
+			MACOSX_BUNDLE_SHORT_VERSION_STRING ${CLAP_WRAPPER_BUNDLE_VERSION}
+			MACOSX_BUNDLE_INFO_PLIST ${CMAKE_SOURCE_DIR}/cmake/VST3_Info.plist.in
+			)
+	add_custom_command(TARGET ${pluginname_vst3} POST_BUILD
```
0ax1, line 116: For Xcode, this generates:
```
set -e
if test "$CONFIGURATION" = "Debug"; then :
  cd /Users/lx/Desktop/wrapper/clap-wrapper/build
  SetFile -a B xyz.vst3
fi
if test "$CONFIGURATION" = "Release"; then :
  cd /Users/lx/Desktop/wrapper/clap-wrapper/build
  SetFile -a B xyz.vst3
fi
if test "$CONFIGURATION" = "MinSizeRel"; then :
  cd /Users/lx/Desktop/wrapper/clap-wrapper/build
  SetFile -a B xyz.vst3
fi
if test "$CONFIGURATION" = "RelWithDebInfo"; then :
  cd /Users/lx/Desktop/wrapper/clap-wrapper/build
  SetFile -a B xyz.vst3
fi
```

and therefore fails to find the `.vst` bundle, as the builds are nested under `Debug`, `Release`, etc. within the `build` directory.

We don't need to branch on the build mode but could use: `SetFile -a B "$CONFIGURATION_BUILD_DIR/xyz.vst3"`

0ax1: A diff issue here seems to be the Xcode annoyance that this step in context of Xcode is just another build step and not strictly a post-build step. This matters for code signing.

It gets defined here:
image

but should be defined under `edit scheme` -> `post-actions` 
image

Otherwise the code-signing fails.

CMakeLists.txt
```diff
@@ -2,15 +2,28 @@
...
+# CLAP_WRAPPER_BUNDLE_VERSION the macOS Bundle Version. Defaults to 1.0
 #
-#	
-
 
 cmake_minimum_required(VERSION 3.20)
 cmake_policy(SET CMP0091 NEW)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum macOS version")
 set(CMAKE_OSX_ARCHITECTURES "x86_64;arm64" CACHE STRING "Build Universal Always")
+set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>") # Statically link the MSVC Runtime
```
0ax1, line 26: For `cmake -GXcode -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCLAP_WRAPPER_OUTPUT_NAME=xyz`,

CMake throws a warning on my end that the `CMAKE_BUILD_TYPE` is not used.

```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_BUILD_TYPE
```

CMakeLists.txt
```diff
@@ -22,110 +35,106 @@ project(clap-wrapper
...
-include(cmake/enable_sdks.cmake)
-
-# ------------------------------------------------------------------------------
 
 set(CMAKE_CXX_STANDARD 17)
-
-# set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Build for 10.15")
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
0ax1, line 41: As I haven't dug deep before into the VST3 CMake files, would you mind dropping a line on why these are necessary / how this has been leading to problems before?

0ax1 (commented):
Fantastic maneuver moving to a simpler VST3 SDK CMake setup!

Couple of initial questions and remarks inline.

CMakeLists.txt
0ax1: Is this simply up to the invocation of the build step? I can see that `-DDEVELOPMENT=1` and `-DRELEASE=1` are set respectively for the diff build modes.

image

0ax1: So the command could actually just be: `cmake -GXcode -B ./build -DCLAP_WRAPPER_OUTPUT_NAME=xyz`. 

Very nice!

CMakeLists.txt
0ax1: The build completes without errors with `Make` though on my end.

CMakeLists.txt
0ax1, line 26: How about adding `set(CMAKE_COLOR_DIAGNOSTICS ON)` to get colored diagnostics for each generator? E.g. `ninja` is otherwise uncolored.

0ax1:
Another thing I noticed, the new CMake doesn't symlink on macOS any longer to `~/Library/Audio/Plug-Ins/VST3/xyz.vst3`.

CMakeLists.txt
baconpaul: Ahh right I should just exclude this in Xcode

CMakeLists.txt
baconpaul: yes for dual build generators you don't need a build type. For single build generators like unixmakefile and ninja you do.

CMakeLists.txt
baconpaul: Sure. So visibility hiding means the .so / .dylib only exports explicitly exported symbols. This matters quite a bit in plugin world on unixes because you end up with all sorts of nasty symbol conflicts.

That is, if you happen to link library abc1.2.3 and export your symbols, you can become a source for abc downstream and a second plugin which dynamically loads it can get them from you rather than the daw. Or if you run two copies of yourself built at different versions in the DAW you can get conflicts

So really best practice on the unixes is to hide symbols so you *only* export the plugin entry point. That's what this does.

I have no idea why the VST3 sdk doesn't do it.But it should.

CMakeLists.txt
baconpaul: Sure it doesn't break CLion so I'll add it to my push!

CMakeLists.txt
baconpaul: Will be in next push.

baconpaul:
OK [c5d5fe1](https://github.com/free-audio/clap-wrapper/pull/63/commits/c5d5fe1c8f491e393485786c1389132331fb065f) responds to review comments here and on discord.

0ax1 (approved):
Thanks for addressing the remarks and providing more context! 

LGTM from my end. Note that I mostly double checked for macOS here.

baconpaul:
Great. I built and ran all 3 platforms and also set up ci for Linux. Appreciate the comments and review!

baconpaul:
> Another thing I noticed, the new CMake setup doesn't symlink on macOS any longer to `~/Library/Audio/Plug-Ins/VST3/xyz.vst3`. Is that intended?

Ahh yes that soft link always broke me when I swapped back to juce which does a proper install so I didn’t port it, but I should add the proper install

0ax1:
> > Another thing I noticed, the new CMake setup doesn't symlink on macOS any longer to `~/Library/Audio/Plug-Ins/VST3/xyz.vst3`. Is that intended?
> 
> Ahh yes that soft link always broke me when I swapped back to juce which does a proper install so I didn’t port it, but I should add the proper install

Personal pref would be to make this an explicit opt-in, rather than a default.

baconpaul:
> > > Another thing I noticed, the new CMake setup doesn't symlink on macOS any longer to `~/Library/Audio/Plug-Ins/VST3/xyz.vst3`. Is that intended?
> > 
> > 
> > Ahh yes that soft link always broke me when I swapped back to juce which does a proper install so I didn’t port it, but I should add the proper install
> 
> Personal pref would be to make this an explicit opt-in, rather than a default.

Yes the way JUCE does it is it has a COPY_AFTER_BUILD which defaults to false

In Surge and our other properties we then have a SURGE_COPY_AFTER_BUILD which defaults to false which if true, well, does a copy after build.

I was thinking I could do that if that sounds OK to you.

CMakeLists.txt
```diff
@@ -22,110 +36,108 @@ project(clap-wrapper
...
+# Support some of the options at cmake time
+target_compile_options(${pluginname_vst3} PRIVATE
 		-DCLAP_SUPPORTS_ALL_NOTE_EXPRESSIONS=$<IF:$<BOOL:${CLAP_SUPPORTS_ALL_NOTE_EXPRESSIONS}>,1,0>
 		)
 
-if (NOT ${CLAP_VST3_TUID_STRING} STREQUAL "")
-	message(STATUS "Using cmake-specified VST3 TUID ${CLAP_VST3_TUID_STRING}")
-	target_compile_options(${pluginname} PRIVATE
+if (NOT "${CLAP_VST3_TUID_STRING}" STREQUAL "")
+	message(STATUS "clap-wrapper: Using cmake-specified VST3 TUID ${CLAP_VST3_TUID_STRING}")
```
defiantnerd, line 90: I'd add code that only the first plugin will have the TUID applied (which is matching the intended case of a single plugin in the library)

CMakeLists.txt
```diff
@@ -22,110 +35,106 @@ project(clap-wrapper
...
+set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 
-# ------------------------------------------------------------------------------
-
-
-if(NOT CMAKE_BUILD_TYPE)
-    message(STATUS "CMAKE_BUILD_TYPE Type Unspecified; picking Release")
-    set(CMAKE_BUILD_TYPE "Release" CACHE STRING "" FORCE)
+if(${CMAKE_SIZEOF_VOID_P} EQUAL 4)
+	message(FATAL_ERROR "Clap Wrapper requires a 64 bit build")
```
defiantnerd: we should not disable 32 bit builds of the wrapper

cmake/enable_sdks.cmake
```diff
@@ -67,45 +67,50 @@ function(DetectVST3SDK)
...
 
   cmake_path(CONVERT "${VST3_SDK_ROOT}" TO_CMAKE_PATH_LIST VST3_SDK_ROOT)
-  message(STATUS "VST3 SDK location: ${VST3_SDK_ROOT}")
+  message(STATUS "clap-wrapper: VST3 SDK location: ${VST3_SDK_ROOT}")
   set(VST3_SDK_ROOT "${VST3_SDK_ROOT}" PARENT_SCOPE)
 
 endfunction()
 
 function(DefineCLAPASVST3Sources)
+	file(GLOB VST3_GLOB
```
defiantnerd, line 76: should we really glob everyhing? we actually don't need all the cpp files.

src/detail/os/windows.cpp
```diff
@@ -117,7 +117,7 @@ namespace os
 		memset(&wc, 0, sizeof(wc));
 		wc.cbSize = sizeof(wc);
 		wc.hInstance = ghInst;
-		wc.lpfnWndProc = &Wndproc;
+		wc.lpfnWndProc = (WNDPROC)&Wndproc;
```
defiantnerd, line 120: did this show a warning?

defiantnerd (approved):
Thanks @baconpaul for the work and @0ax1 for the review.

PR #69, opened 2023-09-01
Build a shell AU with cmake

baconpaul:
If (1) CMAKE_WRAPPER_BUILD_AUV2 is set and (2) we can find the AUDIOUNIT_SDK_ROOT using the same pattern as the rest, then build a stub AUV2.

This AUV2 isn't usable. And to complete this work will require a bunch of information to fill out the info.plist. But it will scan in auval and show up as an audio unit.

So commit this to 'next' as an off-by-default starting point for AUV2


baconpaul:
Closing this - it will be different if current cmake rework ends up in place

Issue #14, opened 2022-10-14
implementation of linux timers


defiantnerd:
Linux timers need implementation. 

- the timers should also be available when the UI is not open
- the timer must not be concurrent to the actual UI/main thread

baconpaul:
I did this in surge 1.9 based on some code in the SDK. I can port thst over here

PR #85, opened 2023-09-02
Linux Timer and FD support (when VST3 Runloop Available)

baconpaul:
This implements linux timer and FD support, and fires the linux onIdle loop in the VST3 host. It accomplishes this by tapping into the Steinberg::Linux::IRunLoop which only exists on the pluginframe. This means absent a UI you currently don't get linxu timer and fd events.  But since the primary use of these features to date is UI updates, it make sa large number of claps work in the vst3 wrapper in reaper and bitwig on initial tests.

Separately from this commit, we should decide what we want to do about FD polling and timers absent a UI, probably by consulting with the bitwig and reaper native clap implementations.

Addresses (and maybe closes) #14


src/wrapasvst3_entry.cpp
```diff
@@ -84,8 +83,7 @@ bool findPlugin(Clap::Library& lib, const std::string& pluginfilename)
 		// try to find it the CLAP folder immediately
 		auto k1 = i / pluginfilename;
 		LOGDETAIL("scanning for binary: {}", k1.u8string().c_str());
-		std::cout << "scanning for " << k1 << std::endl;
```
defiantnerd, line 87: oh, right - thanks!

PR #87, opened 2023-09-02
Fix a small warning on param info access

baconpaul:
Vst3Parameeter::create used to do

if (info->module && info->module[0])

but clap_param_info_t defines the module a `char module[CLAP_PATH_SIZE] so it is always a realized pointer. This made macos throw the warning

address of array 'info->module' will always evaluate to 'true' [-Wpointer-bool-conversion]

correctly. So just fix it!


defiantnerd:
Just a leftover from defensive  programming strategies. :)

baconpaul:
yeah! good instinct! I checked if the compiler was correct 3 times before changing :)

PR #88, opened 2023-09-02
Add an 'in-bundle' self contained load macOS option

baconpaul:
macOS bundles support a PlugIn directory (Contents/PlugIns) which contain subordinate plug-ins to be loaded. This commit does two things

1. Adds a MACOS_EMBEDDED_CLAP_LOCATION option to target_add_auv2/vst3_wrapper which if set creates the Contents/Plugins in the bundle and copys the plugin there; and

2. Modifies the mac system paths to look up the current running bundle and then use CFBundleCopyBuiltInPlugInsURL to get this location and pre-pend it to the clap load search path.

Result: Fully self contained VST3 and AUV2.


cmake/enable_sdks.cmake
```diff
@@ -337,6 +339,15 @@ function(target_add_vst3_wrapper)
 					WORKING_DIRECTORY $<TARGET_PROPERTY:${V3_TARGET},LIBRARY_OUTPUT_DIRECTORY>
 					COMMAND SetFile -a B "$<TARGET_PROPERTY:${V3_TARGET},MACOSX_BUNDLE_BUNDLE_NAME>.$<TARGET_PROPERTY:${V3_TARGET},BUNDLE_EXTENSION>")
 		endif()
+
+		if (NOT ${V3_MACOS_EMBEDDED_CLAP_LOCATION} STREQUAL "")
+			add_custom_command(TARGET ${V3_TARGET} POST_BUILD
+					WORKING_DIRECTORY $<TARGET_PROPERTY:${V3_TARGET},LIBRARY_OUTPUT_DIRECTORY>
+					COMMAND ${CMAKE_COMMAND} -E echo "Installing ${V3_MACOS_EMBEDDED_CLAP_LOCATION} in $<TARGET_PROPERTY:${V3_TARGET},MACOSX_BUNDLE_BUNDLE_NAME>.$<TARGET_PROPERTY:${V3_TARGET},BUNDLE_EXTENSION>/Contents/PlugIns/$<TARGET_PROPERTY:${V3_TARGET},MACOSX_BUNDLE_BUNDLE_NAME>.clap"
+					COMMAND ${CMAKE_COMMAND} -E make_directory "$<TARGET_PROPERTY:${V3_TARGET},MACOSX_BUNDLE_BUNDLE_NAME>.$<TARGET_PROPERTY:${V3_TARGET},BUNDLE_EXTENSION>/Contents/PlugIns"
```
defiantnerd, line 347: I like this a lot!!

PR #91, opened 2023-09-04
Changes to allow to work as subordinate cmake

baconpaul:
These changes allow the cmake functions to work at a not-top-level and to supress the default targets so you can augment your targets with wrapitude.

The changes mostly amount to

1. CACHE STRING vs PARENT_SCOPE. Some thing sjust need a lifetime beyond the parent scope when PARENT_SCOPE is not top level.
2. The functions called from anywhere mean we can't rely on the working directory being the current soruce, so cache current source and use it in a couple of key places.
3. Add an option to skip the build or activate it, with them on by default. I chose a "DONT ADD" rather than "ADD" flag here but might change it. This also pushes us to cmake 3.21 for PROJECT TOP LEVEL
4. Set visibilty flags for objective c as well as c++


baconpaul:
**PLEASE HOLD OFF ON REVIEW AND MERGE FOR NOW**. I'm pushing this up to test CI but want to do some more manual win/lin testing today.

baconpaul:
OK this is ready now. Tested on all three and it builds properly both as lead and subordinate.

defiantnerd:
Looks all good for me.

Issue #86, opened 2023-09-02
Implement toString / fromString on VST3 Parameter


baconpaul:
IThe wrapper doesn't call `param::value_to_text` anywhere or `text_to_value`

So what we need to do is on Vst3Parameter add an overload of

```
    virtual void toString (ParamValue valueNormalized, String128 string) const;
    virtual bool fromString (const TChar* string, ParamValue& valueNormalized) const;
```

then implement `to_string` as `_pluigin->_paramextension->value-to_text`.

Since the VST3 Parameter object doesn't have a reference to the plugin, the idomatic way seems to be as follows

1. Add a pair of `std::functions` to the VST3 Parameter `std::function<std::string(float)> valueToString` and `std::function<float(const std::string &)> valueFromString`
2. Call those in the overloads with a null check. Deal with the TCHar/Widechar thing on windows properly
3. Have them implemented to bind the plugin and clap id. Have them call through to the to string from string ids
4. Test etc

defiantnerd:
implemented in IEditController overloads.

PR #94, opened 2023-09-05
fixes https://github.com/free-audio/clap-wrapper/issues/86

defiantnerd:
uses the CLAP plugin to render string representations of parameter values and tries to apply values given as strings to the CLAP plugin parameter.


defiantnerd:
I will merge despite the unsuccesfull check and we take care of it from next

PR #97, opened 2023-09-06
Cmake 230906

defiantnerd:
change the library name so the targets are easier identifiable. added some comments to explain the mechanics.


defiantnerd:
needs to be merged to next.

PR #99, opened 2023-09-06
clap wrapper target lib rename changes continued

baconpaul:
1. Fix a target which had missed the rename sweep (due to merge order difference)
2. Rename the AUv2 targets symmetrically with the VST3 ones


defiantnerd:
Thanks, Paul

PR #104, opened 2023-09-07
Implement outbound CLAP_EVENT_NOTE_* from wrapped plugins

baconpaul:
Push them as kNoteOn/Off events onto the result bus.


src/detail/vst3/process.cpp
```diff
@@ -693,6 +733,7 @@ namespace Clap
     case CLAP_EVENT_MIDI:
```
defiantnerd: We will have to map those to the IMidiMapping parameters..

baconpaul: Yes I opened issue #105 to remember taht

defiantnerd:
Thank you.

PR #106, opened 2023-09-07
Set kIsList to true if CLAP_PARAM_IS_STEPPED

### LINKED ISSUE #89: Parameters that represent a list should get the kIsList flag
we need to check how much of the ParameterFlags for VST3 can be determined from the CLAP parameter definition:

```c++
enum ParameterFlags
  {
    kNoFlags = 0,		///< no flags wanted
    kCanAutomate = 1 << 0,	///< parameter can be automated
    kIsReadOnly = 1 << 1,	///< parameter cannot be changed from outside the plug-in (implies that kCanAutomate is NOT set)
    kIsWrapAround = 1 << 2,	///< attempts to set the parameter value out of the limits will result in a wrap around [SDK 3.0.2]
    kIsList = 1 << 3,	///< parameter should be displayed as list in generic editor or automation editing [SDK 3.1.0]
    kIsHidden = 1 << 4,	///< parameter should be NOT displayed and cannot be changed from outside the plug-in 
    ///< (implies that kCanAutomate is NOT set and kIsReadOnly is set) [SDK 3.7.0]

    kIsProgramChange = 1 << 15,	///< parameter is a program change (unitId gives info about associated unit 
    ///< - see \ref vst3ProgramLists)
    kIsBypass = 1 << 16	///< special bypass parameter (only one allowed): plug-in can handle bypass
    ///< (highly recommended to export a bypass parameter for effect plug-in)
  };
```

baconpaul:
Closes #89


baconpaul:
@defiantnerd wanted to go a different route here it seems so let me close this. It will still link to the above issue in case this is the path we take.

Issue #81, opened 2023-09-02
Ramp up compiler warnings


0ax1:
As of now, the `clap-wrapper` target uses the default minimum of compiler warnings. In order to catch as many errors as possible at compile time, additional warning groups such as `-Wall`, `-Wextra` and `-Wpedantic` should be considered.

After resolving all warnings for the added groups, we could think about enabling warnings as errors: `-Werror`.

0ax1:
landed for clang/gcc in: https://github.com/free-audio/clap-wrapper/pull/109

baconpaul:
we are now `-Wall -Wextra -Wpedantic -Werror` on gcc and clang. Thanks for raising this.

I suggest we open a separate issue for MSVC and close this one when we make 0.5.0

PR #110, opened 2023-09-08
Elevate warnigns to errors on gcc and clang

baconpaul:
Elevate warnigns to error on gcc and clang.
This sets our error level pretty high. CI will be an important part of this strategy working.


baconpaul:
OK so @defiantnerd this makes the vst3 wrapper use all the warnings on mac and linux.

I think we should turn this on but it will mean broken CI sometimes since they warnings differ by platform. If you are comfy with that I am also. But I can see wanting to wait.

It would close #81 if we did it though, or at least close that then encourage someone to do same for MSVC.

PR #114, opened 2023-09-09
AUv2 PList Generator - step one

baconpaul:
The AUv2 PList generator needs to be way smarter than single entry in cmake, and this commit starts us down that path. The commit creates a runtime-binary creator which generates the plist and associated c++ needed to make entry points, descriptions, etc...

Along the way, it also refactors the detail library moving some os stuff to the VST3 and making the clap specific detail library available to targets other than the VST3


baconpaul:
I still have a few checks I want to do here; this is to test CI. Please hold off until I comment here.

baconpaul:
This is now ready for review

baconpaul:
Oh one comment. I moved *all* of `details/os` to `details/vst3/os`

I am sure when we finish AUv2 we will find `details/os` used to contain part vst3 specific and part not

So I assume we will end up with `details/shared/os` as a separate set of code which factors out bits.

I did *not* do that with this commit. I just made sure the vst3 stuff was out of the way of the clap stuff for inclusion in the build helper.

PR #120, opened 2023-09-10
Add an AUv2 factory option for subtype etc...

baconpaul:
The subtype etc can be determined bu either cmake arguments, a hash of the id, or explicit sets of optiosn through a factor akin to the vst3 factory. This commit implements the latter.

Once merged, conduit will use this to explicitly set its vst3 and auv2 options.

### DIFF
--- a/src/clap_proxy.cpp
+++ b/src/clap_proxy.cpp
@@ -239,6 +239,8 @@ namespace Clap
     {
       _parentHost->setupParameters(_plugin, _ext._params);
     }
+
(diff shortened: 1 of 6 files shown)


src/clap_proxy.cpp
```diff
@@ -239,6 +239,8 @@ namespace Clap
     {
       _parentHost->setupParameters(_plugin, _ext._params);
     }
+
+    _parentHost->setupWrapperSpecifics(_plugin);
```
defiantnerd, line 243: ouch :)

PR #122, opened 2023-09-10
Enable consistent clang format for the project

baconpaul:
1. Add a .clang-format which is basically 2 spaces, allman braces, embedded namespaces.
2. Reformat the entire codebase with find src -iname "*.cpp" -o -iname "*.h" -o -iname "*.mm" | xargs clang-format -i
3. add a cmake code checks target which currently just checkes that the code remains clang formatted
4. Add a github action which runs that code check target on ubuntu on a PR

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -14,7 +14,7 @@ BreakConstructorInitializers: BeforeComma
 BreakBeforeBraces: Allman
 ColumnLimit: 105 
 SortIncludes: false
-NamespaceIndentation: All
+NamespaceIndentation: Inner
 ReflowComments: false
 AllowShortFunctionsOnASingleLine: None
 ---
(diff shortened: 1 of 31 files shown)


.clang-format
```diff
@@ -0,0 +1,25 @@
+﻿---
+UseTab: Never
+---
+Language: Cpp
+BasedOnStyle: Google
+IndentWidth: 2
+AlignAfterOpenBracket: Align
+BreakBeforeBraces: Allman
+ColumnLimit: 250
```
0ax1, line 9: Do we really wanna turn off automatic line breaks? Kinda hard to work with on a laptop. How about something generous like 105?

0ax1 (approved):
🚀

Issue #118, opened 2023-09-09
Various validator issues (VST3)


baconpaul:
Using this as an issue for various validator problems.

Two validators in question

1. The Steinberg VST3 validator which ships with the SDK
2. The tracktion pluginval https://github.com/Tracktion/pluginval

This report is building conduit at  44dec6fa15108 which has clap-wrapper at  dd5e512bc

## steinberg validator: Unit Info problems

Conduit params have modules (groups). So the conduit validation shows for the poly synth (correctly)

```
[Scan Parameters]
Info:  ===Scan Parameters ====================================
Info:  This component exports 10 parameter(s)
Info:     Parameter 000 (id=1378): [title="Unison Count"] [unit=""] [type = D, default = 0.333333, unit = 1]
Info:     Parameter 001 (id=2391): [title="Unison Spread"] [unit=""] [type = F, default = 0.100000, unit = 1]
Info:     Parameter 002 (id=8675309): [title="Unison Detune"] [unit=""] [type = F, default = 0.500000, unit = 1]
Info:     Parameter 003 (id=2874): [title="Amplitude Attack"] [unit=""] [type = F, default = 0.050000, unit = 2]
Info:     Parameter 004 (id=728): [title="Amplitude Release"] [unit=""] [type = F, default = 0.050000, unit = 2]
Info:     Parameter 005 (id=1942): [title="Bypass Amp Envelope"] [unit=""] [type = D, default = 0.000000, unit = 2]
Info:     Parameter 006 (id=17): [title="Cutoff"] [unit=""] [type = F, default = 0.539683, unit = 3]
Info:     Parameter 007 (id=94): [title="Resonance"] [unit=""] [type = F, default = 0.707107, unit = 3]
Info:     Parameter 008 (id=87612): [title="PreFilter VCA"] [unit=""] [type = F, default = 1.000000, unit = 3]
Info:     Parameter 009 (id=14255): [title="Filter Type"] [unit=""] [type = D, default = 0.000000, unit = 3]
Info:  No bypass parameter found. This is an instrument.
[Succeeded]
```

but then shortly thereafter

```
[Scan Units]
Info:  ===Scan Units ====================================
Info:  This component has 4 unit(s).
ERROR: Unit 000: Invalid ID!
[XXXXXXX Failed]
```

## Wrong Thread with pluginval

[...]

PR #128, opened 2023-09-11
Set up the VST3 Root Until so vst3 validator succeeds

baconpaul:
Set up all the IDs required to get a valid root unit in the steinberg validator. Addresses #118


defiantnerd:
Thanks!

PR #129, opened 2023-09-11
Refactor the cmake compile options a bit

baconpaul:
A few options (like -DPLATFORM=1 and macos filesystem and warnings) were starting to be applied multiple places and not consistently. For instance I had forgotten them on the auv2 build helper. So

1. Make a `clap-wrapper-compile-options` interface which sets up all these things
2. Set it as a link dep of the things which need it
3. Fix the build bugs with the auv2 helper from turning on warnings there and
4. Add an option to likn with asan, which is still a work in progress

### DIFF
--- a/cmake/enable_sdks.cmake
+++ b/cmake/enable_sdks.cmake
@@ -258,10 +258,19 @@ if (APPLE)
 endif()
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
 	target_compile_options(clap-wrapper-compile-options INTERFACE -Wall -Wextra -Wno-unused-parameter -Wpedantic -Werror)
-	if (${CLAP_WRAPPER_ENABLE_ASAN})
-		message(STATUS "clap-wrapper: enabling asan build")
-		target_compile_options(clap-wrapper-sanitizer-options INTERFACE -fsanitize=address -fsanitize=undefined)
-		target_link_options(clap-wrapper-sanitizer-options INTERFACE -fsanitize=address -fsanitize=undefined)
+	if (${CLAP_WRAPPER_ENABLE_SANITIZER})
+		message(STATUS "clap-wrapper: enabling sanitizer build")
+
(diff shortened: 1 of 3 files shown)


cmake/enable_sdks.cmake
```diff
@@ -247,6 +248,25 @@ if (APPLE)
...
+if (APPLE)
+	target_link_libraries(clap-wrapper-compile-options INTERFACE macos_filesystem_support)
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
+	target_compile_options(clap-wrapper-compile-options INTERFACE -Wall -Wextra -Wno-unused-parameter -Wpedantic -Werror)
+	if (${CLAP_WRAPPER_ENABLE_ASAN})
+		message(STATUS "clap-wrapper: enabling asan build")
+		target_compile_options(clap-wrapper-sanitizer-options INTERFACE -fsanitize=address -fsanitize=undefined)
+		target_link_options(clap-wrapper-sanitizer-options INTERFACE -fsanitize=address -fsanitize=undefined)
+		target_link_libraries(clap-wrapper-compile-options INTERFACE clap-wrapper-sanitizer-options)
```
0ax1, line 265: Just to double check, this did work with the XCode generator?

0ax1 (approved):
lgtm!

one nit inline.

cmake/enable_sdks.cmake
```diff
@@ -247,6 +248,25 @@ if (APPLE)
...
+	endif()
+endif()
+
+
```
0ax1, line 269: just to highlight, there's more options we could enable:
`-fsanitize=address,undefined,float-divide-by-zero -fno-sanitize-recover -fsanitize-address-use-after-return=always -fsanitize-address-use-after-scope`

baconpaul: fno-sanitize-recover suppresses call stacks in some situations but agree with rest! Will add.

Issue #127, opened 2023-09-11
VST3 validator crash - Output queue missing - ASAN


0ax1:
For one of the VST3 validator tests, no output queue is provided, such that the wrapper crashes on https://github.com/free-audio/clap-wrapper/blob/next/src/detail/vst3/process.cpp?plain=1#L701 with `_vstdata->outputParameterChanges` being `nullptr`. This report/backtrace was generated when running the validator and wrapper compiled with [ASAN](https://clang.llvm.org/docs/AddressSanitizer.html).

Backtrace:
```
#0	0x0000000106730538 in Clap::ProcessAdapter::enqueueOutputEvent(clap_event_header const*) at clap-wrapper/src/detail/vst3/process.cpp:701
#1	0x00000001067248d0 in Clap::ProcessAdapter::output_events_try_push(clap_output_events const*, clap_event_header const*) at clap-wrapper/src/detail/vst3/process.cpp:456
#2	0x00000001062ace94 in apply_parameter_changes_to_engine at Clap/Clap.m:77
#3	0x00000001062ac96c in plug_process at Clap/Clap.m:558
#4	0x000000010672998c in Clap::ProcessAdapter::process(Steinberg::Vst::ProcessData&) at clap-wrapper/src/detail/vst3/process.cpp:397
#5	0x0000000106758e60 in ClapAsVst3::process(Steinberg::Vst::ProcessData&) at clap-wrapper/src/wrapasvst3.cpp:109
#6	0x00000001002c8c70 in Steinberg::Vst::ProcessTest::run(Steinberg::ITestResult*) at public.sdk/source/vst/testsuite/processing/process.cpp:95
#7	0x000000010000e0cc in Steinberg::Vst::Validator::runTestSuite(Steinberg::Vst::TestSuite*, char const*) at public.sdk/samples/vst-hosting/validator/source/validator.cpp:742
#8	0x000000010000ee3c in Steinberg::Vst::Validator::runTestSuite(Steinberg::Vst::TestSuite*, char const*) at public.sdk/samples/vst-hosting/validator/source/validator.cpp:802
#9	0x000000010000ee3c in Steinberg::Vst::Validator::runTestSuite(Steinberg::Vst::TestSuite*, char const*) at public.sdk/samples/vst-hosting/validator/source/validator.cpp:802
[...]

PR #130, opened 2023-09-11
Fix issue 127 (crash in validator)

defiantnerd:
checking now if there is an output queue object,
fixing https://github.com/free-audio/clap-wrapper/issues/127


baconpaul:
Lgtm

PR #137, opened 2023-09-14
fixing mac dynamic clap loader

defiantnerd:
dynamic loading of CLAPs didn't work.


0ax1 (approved):
works on my machine™

baconpaul:
ha i had never actually tried it I realized; i only build the auv2 as conduit and made sure csd compiled and validated but not that it loaded.  so this is even a more insidious "kinda looks like it might work" :)

Exciting stuff

PR #138, opened 2023-09-14
A rudimentary standalone wrapper

baconpaul:
This commit provides a rudimentary standalone wrapper. The wrapper makes the following choices

1. It uses rtaudio/rtmidi for auido and midi support
2. It is a Cocoa (mac) GTK3 (lin) and just-exe-until-timo-gets-it (win) applications
3. The application binds all midi and default audio at 48k with no user choices
4. On mac and lin the application shows the gui if present
5. The commit adds cmake rules to decorate an exeecutable as as standalone with clap_entry either statically linked

There are myriad things to do to make this a production worth standalone, including state, saving options, choosing audio, a win gui, a timer loop on linux which isn't laughable, an fd loop on linux, and much more. But this at least puts the stake in the ground about project cmake and directory structure and actuall does work with conduit and clap saw demo.

### DIFF
--- a/src/detail/standalone/macos/AppDelegate.mm
+++ b/src/detail/standalone/macos/AppDelegate.mm
@@ -1,0 +1,120 @@
+#import "AppDelegate.h"
+
+#include <AVFoundation/AVFoundation.h>
+
+#include "detail/standalone/entry.h"
+#include "detail/standalone/standalone_details.h"
+
+#include "detail/clap/fsutil.h"
+
+@interface AppDelegate ()
+
+@property(assign) IBOutlet NSWindow *window;
+@end
+
+@implementation AppDelegate
+
+- (void)applicationDidFinishLaunching:(NSNotification *)aNotification
+{
+  // Insert code here to initialize your application
+  const char *argv[2] = {OUTPUT_NAME, 0};
+
+  const clap_plugin_entry *entry{nullptr};
+#ifdef STATICALLY_LINKED_CLAP_ENTRY
+  extern const clap_plugin_entry clap_entry;
+  entry = &clap_entry;
+#else
+  // Library shenanigans t/k
+  std::string cn{HOSTED_CLAP_NAME};
+  LOG << "Loading " << cn << std::endl;
+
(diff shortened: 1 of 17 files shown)


src/detail/standalone/macos/AppDelegate.mm
```diff
@@ -0,0 +1,120 @@
...
+
+  if (!entry)
+  {
+    return;
+  }
+
+  std::string pid{PLUGIN_ID};
+  int pindex{PLUGIN_INDEX};
+
+  switch ([AVCaptureDevice authorizationStatusForMediaType:AVMediaTypeVideo])
```
baconpaul, line 55: ahh this should be type audio

src/detail/standalone/standalone_host.cpp
```diff
@@ -0,0 +1,256 @@
...
+  sh->runTimerFn();
+  return true;
+}
+
+bool StandaloneHost::register_timer(uint32_t period_ms, clap_id *timer_id)
+{
+  if (!timersStarted)
+  {
+    LOG << "Stargint Timers" << std::endl;
+    g_timeout_add(30, time_handler, this);
```
baconpaul, line 199: this needs to be in a CLAP_WRAPPER_HAS_GTK3 guard

0ax1 (approved):
lgtm as a first throw.

one nit: as noted inline, maybe worth considering being a little more descriptive in some instances for variable names.

src/detail/standalone/macos/AppDelegate.mm
```diff
@@ -0,0 +1,120 @@
...
+  // Library shenanigans t/k
+  std::string cn{HOSTED_CLAP_NAME};
+  LOG << "Loading " << cn << std::endl;
+
+  auto pts = Clap::getValidCLAPSearchPaths();
+
+  auto lib = Clap::Library();
+
+  for (const auto &p : pts)
+  {
```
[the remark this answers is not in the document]
baconpaul: Oh sure I’ll update that 

bad habit that I give super local locals cruddy names

PR #139, opened 2023-09-14
Breakup cmake

baconpaul:
This PR breaks up the cmake into more manageable chunks and factors lots of things into guarantee_ methods to make sure preconditions are met only when needed. it also removes a lot of the CACHE SET stuff we had to do for subordinate cmake. And generally makes things a bit easier to find I think.

And adds a Xcode stage

I'm still pondering whether this is right or finished. So I'm not proposing we merge it until I get another chance to think about it and look tomorrow. But also welcome your thoughts. (Also I want to see if CI works)


.github/workflows/pullreq.yml
```diff
@@ -75,3 +75,28 @@ jobs:
...
+        os: [macos-latest]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Build project
+        run: |
+          cmake -S . -B ./build -GXcode -DCMAKE_BUILD_TYPE=Debug -DCLAP_WRAPPER_DOWNLOAD_DEPENDENCIES=TRUE -DCLAP_WRAPPER_BUILD_AUV2=TRUE  -DCLAP_WRAPPER_OUTPUT_NAME=downloadplug
```
0ax1, line 94: 👍

cmake/base_sdks.cmake
```diff
@@ -0,0 +1,286 @@
...
+    set(RES "")
+
+    message(STATUS "clap-wrapper: searching for '${SEARCH_SDKDIR}' in \"${CMAKE_CURRENT_SOURCE_DIR}\"...")
+
+    if ( RES STREQUAL "" AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/libs/${SEARCH_SDKDIR}")
+        set(RES "${CMAKE_CURRENT_SOURCE_DIR}/libs/${SEARCH_SDKDIR}")
+        message(STATUS "clap-wrapper: '${SEARCH_SDKDIR}' detected in libs subdirectory")
+    endif()
+
+    if ( RES STREQUAL "" AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${SEARCH_SDKDIR}")
```
0ax1, line 16: `elseif`?

cmake/base_sdks.cmake
```diff
@@ -0,0 +1,286 @@
+# locates certain paths/files in specific locations
+# used to locate the base paths for the different Plugin SDKs
+function(search_for_sdk_source)
+    set(oneValueArgs SDKDIR RESULT)
+    cmake_parse_arguments(SEARCH "" "${oneValueArgs}" "" ${ARGN} )
+
+    set(RES "")
```
0ax1, line 7: Can this one be dropped if we handle all cases after the initial `if` with `elseif`?

cmake/base_sdks.cmake
```diff
@@ -0,0 +1,286 @@
...
+        message(STATUS "clap-wrapper: '${SEARCH_SDKDIR}' detected in libs subdirectory")
+    endif()
+
+    if ( RES STREQUAL "" AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${SEARCH_SDKDIR}")
+        set(RES "${CMAKE_CURRENT_SOURCE_DIR}/${SEARCH_SDKDIR}")
+        message(STATUS "clap-wrapper: '${SEARCH_SDKDIR}' SDK detected in subdirectory")
+    endif()
+
+    if ( RES STREQUAL "" AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/../${SEARCH_SDKDIR}")
+        set(RES "${CMAKE_CURRENT_SOURCE_DIR}/../${SEARCH_SDKDIR}")
```
0ax1, line 22: In CMake and `if` doesn't introduce scope right? Then we could still print `res` subsequently as part of the `message` call.

0ax1 (commented):
Nice breakup into smaller components!

Given that CMake is quite idiosyncratic, I'm wondering whether adding a bit more inline comments/docs can help to improve readability. Having been absent from CMake myself for about 3 years, some of the CMake patterns are really not obvious to me

cmake/base_sdks.cmake
baconpaul: yes

cmake/base_sdks.cmake
baconpaul: yes

baconpaul:
> Nice breakup into smaller components!
> 
> Given that CMake is quite idiosyncratic, I'm wondering whether adding a bit more inline comments/docs can help to improve readability. Having been absent from CMake myself for about 3 years, some of the CMake patterns are really not obvious to me

100%

I will add some more extensive comments tomorrow am.

baconpaul:
OK Alex I just pushed a diff which responds to your review but also adds some comments on what's broadly going on.

cmake/wrap_standalone.cmake
```diff
@@ -0,0 +1,134 @@
...
+
+            MACOS_EMBEDDED_CLAP_LOCATION
+            )
+    cmake_parse_arguments(SA "" "${oneValueArgs}" "" ${ARGN} )
+
+    if (NOT DEFINED SA_TARGET)
+        message(FATAL_ERROR "clap-wrapper: target_add_standalone_wrapper requires a target")
+    endif()
+
+    if (NOT TARGET ${SA_TARGET})
```
0ax1, line 22: Can `${SA_TARGET}` ever be defined if `SA_TARGET` is not defined?

0ax1 (approved):
Looks great, and thanks for adding the docs!

Couple more remarks inline. One general question, is there anything here which could make it difficult in the future having multiple plugins?

cmake/base_sdks.cmake
```diff
@@ -0,0 +1,287 @@
...
+    endif()
+
+    # Just for consistency
+    if (NOT TARGET base-sdk-clap)
+        add_library(base-sdk-clap ALIAS clap)
+    endif()
+endfunction(guarantee_clap)
+
+function(guarantee_vst3sdk)
+    if (NOT TARGET base-sdk-vst3)
```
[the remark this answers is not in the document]
baconpaul: it is needed. I'll add a comment explaining why but basically we call guarantee late in the target_add. Those target_add functions can be called multiple times. So if we are lazy and late someone needs to do the if not. I chose to put it in guarantee so guarantee is safe to call without a check externally.

I will comment this in base_sdks.cmake and repost.

cmake/shared_prologue.cmake
```diff
@@ -0,0 +1,173 @@
...
+        target_link_libraries(clap-wrapper-shared-detail PRIVATE macos_filesystem_support)
+    endif()
+endfunction(guarantee_clap_wrapper_shared)
+
+# add a SetFile POST_BUILD for bundles if you aren't using xcode
+function(macos_bundle_flag)
+    set(oneValueArgs TARGET)
+    cmake_parse_arguments(MBF "" "${oneValueArgs}" "" ${ARGN})
+
+    if (NOT APPLE)
```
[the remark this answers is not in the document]
baconpaul: Great point; I added a WARNING not a FATAL_ERROR to help future humans who set up cross compilation

cmake/wrap_standalone.cmake
baconpaul: No but if SA_TARGET is not defined the above fatal error will shut you down.

This checks that, if it is defined, it is a target and I wanted segregated error messages for the different cases.

baconpaul:
>  One general question, is there anything here which could make it difficult in the future having multiple plugins?

Multiple plugins work right now. Conduit actually does it building conduit and the clap saw demo standalone wrapper in one cmake in my dev env. That's also why the guarantee pattern is lazy, as I'll add to the docs now.

PR #144, opened 2023-09-16
XCode Tweaks for AUV2

baconpaul:
1. Strip the wierd non-working signature xcode puts on the intermediate geneating exe
2. Set the PRODUCT_BUNDLE_IDENTIFIER same as the bundle identifier with a cmake xcode extension

### DIFF
--- a/cmake/wrap_auv2.cmake
+++ b/cmake/wrap_auv2.cmake
@@ -49,13 +49,6 @@ function(target_add_auv2_wrapper)
             )
     set(bhtgoutdir "${CMAKE_CURRENT_BINARY_DIR}/${AUV2_TARGET}-build-helper-output")
 
-    if (NOT ${CMAKE_GENERATOR} STREQUAL "Xcode")
-        add_custom_command(TARGET ${bhtg} POST_BUILD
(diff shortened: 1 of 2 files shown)


cmake/wrap_auv2.cmake
```diff
@@ -71,6 +72,7 @@ function(target_add_auv2_wrapper)
                 POST_BUILD
                 WORKING_DIRECTORY ${bhtgoutdir}
                 BYPRODUCTS ${bhtgoutdir}/auv2_Info.plist ${bhtgoutdir}/generated_entrypoints.hxx
+                COMMAND codesign -s - -f "$<TARGET_FILE:${bhtg}>"
```
0ax1, line 75: If we manually do codesigning as part of the build, we should add the option to let the user pass a specific `CODE_SIGN_IDENTITY`. This is relevant when creating installers.

baconpaul: in general i agree but this is not a shipping binary; this is an internal binary which is used to generate the plist and never sent to anyone.

so in this case we just want it to run at build time and Xcode 14 will put a broken signature on it so we don't ever get a completed build.

0ax1 (approved):

PR #151, opened 2023-09-17
feat: priority queue sort

0ax1:


0ax1:
Closing as we concluded that modern C++ implementations of `std::sort` do not heap allocate.  They are therefore realtime safe. Benchmarking further revealed that `std::sort` is significantly faster than sorting through inserting into a priority queue.

Issue #150, opened 2023-09-16
optimize lock-free queue


0ax1:
our current lock-free queue uses the modulo `%` operator which is computationally expensive. whilst compilers as part of optimization can replace `%` with a bitwise `&` for queue sizes which are a multiple of 2, we shouldn't rely on the compiler to do this optimization but use a bitwise `&` explicitly.

0ax1:
addressed in: https://github.com/free-audio/clap-wrapper/pull/154

defiantnerd:
Merged to 0.6.0

PR #154, opened 2023-09-18
perf: swap in `&` for modulo in spsc

0ax1:
see: https://github.com/free-audio/clap-wrapper/issues/150


baconpaul:
shouldn't that *really* be `& (Q-1)`? 

`& Q` where Q is a power of 2 will basically mask out everything.

baconpaul (changes requested):
See comment. I think you mean `& (Q-1)`

0ax1:
> See comment. I think you mean `& (Q-1)`

Yes good catch. The mask ofc needs to be -1.

0ax1:
> > See comment. I think you mean `& (Q-1)`
> 
> Yes good catch. The mask ofc needs to be -1.

fixed

src/detail/standalone/standalone_details.h
```diff
@@ -30,13 +30,16 @@ class fixedqueue
...
+    _tail = (_tail + 1) & _wrapMask;
     return true;
   }
 
  private:
   T _elements[Q] = {};
   std::atomic_uint32_t _head = 0u;
   std::atomic_uint32_t _tail = 0u;
+
+  uint32_t _wrapMask = Q - 1;
```
baconpaul, line 42: since Q is a template argument shouldn't  this be `static constexpr uint32_t` rather than `uint32_t`?

Also we really need to consolidate this into a clap shared target. Fine to merge the dup still but lets make sure we add an issue for that.

baconpaul (changes requested):

src/detail/standalone/standalone_details.h
0ax1: switched to constexpr

> Also we really need to consolidate
100%

baconpaul (approved):

defiantnerd (approved):

PR #156, opened 2023-09-18
Change the way we make a .clap etc... a bundle

baconpaul:
We were creating a bundle with 'SetFile -a B' which adds a finderinfo attribute which blocks signing. The more correct method is to add a PkgInfo

This diff uses a basic BNDL???? PkgInfo and with non-xcode generators copies it into the bundle. With xcode generator it sets the XCODE_ATTRIBUTE_GENERATE_PKGINFO_FILE property to "YES" and lets Xcode do the hard(ish) work. (Presumably of also copying those 8 bytes into a file)


defiantnerd (approved):
Wohoo!

PR #160, opened 2023-09-19
str8ToStr16 overflow warning

baconpaul:
In some cases, Steinberg::str8ToStr16(a,b,N) writes to a[N] which means you need A to have size N+1 or N to be sizeof(a)-1. I chose the later here. Also add gcc-11 (which flags this) to our CI matrix.


0ax1 (approved):
Nice catch! But also.. 🤦‍♀️

PR #162, opened 2023-09-23
Cpp20 and pkginfo tweaks

jatinchowdhury18:
Three changes here:
1. Adjusting the location for the generated MainMenu.nib for the MacOS Standalone target to use `TARGET_FILE_DIR`.
2. Adding `-fno-char8_t` and `/Zc:char8_t-` flags as appropriate to avoid errors between `std::string` and `std::u8string` with C++20.
3. Using `fmt::vformat_to` rather than `fmt::format_to` to avoid errors about the format string not being known at compile-time (see [this issue](https://github.com/fmtlib/fmt/issues/2931))

Particularly for items (2) and (3), I'm not sure that my solution is the best approach, but it seems to be working for me at the moment. If folks have better suggestions, I'm happy to do those instead!

### DIFF
--- a/cmake/wrap_auv2.cmake
+++ b/cmake/wrap_auv2.cmake
@@ -47,6 +47,7 @@ function(target_add_auv2_wrapper)
             "-framework Foundation"
             "-framework CoreFoundation"
             )
+    target_compile_options(${bhtg} PRIVATE -fno-char8_t)
     set(bhtgoutdir "${CMAKE_CURRENT_BINARY_DIR}/${AUV2_TARGET}-build-helper-output")
 
     add_custom_command(TARGET ${bhtg} POST_BUILD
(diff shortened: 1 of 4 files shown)


baconpaul:
Looks correct to me! @defiantnerd wanna peek or just press?

baconpaul:
Oh and solution 2 is the best approach for 20. With 23 a person may want it off I understand but we could use generstor expresssions then

defiantnerd:
Thanks!

jatinchowdhury18:
> Oh and solution 2 is the best approach for 20. With 23 a person may want it off I understand but we could use generstor expresssions then

Makes sense. Luckily all the instances where that issue popped up were in .cpp files or "private" header files, meaning I was able to make those flags `PRIVATE` everywhere I added them, so they shouldn't affect user code anyway.

PR #165, opened 2023-09-25
target config property in post build step

defiantnerd:
Adding the target config property to the post build copy instruction.
This way, Xcode can copy this properly to the plugin folder.


baconpaul:
This will break single config generators

Lemme take a look this morning - there’s a gen ex for that subdir or the .. strategy may work

PR #166, opened 2023-09-25
MacOS COPY_AFTER_BUILD correct for multi-config generators

baconpaul:
Yet again the lack of a gen exp for the top of the bundle proves awkward. But th TARGET_FILE_DIR/.. patterns are proving best.


defiantnerd (approved):
Yeah - that'll work better for both generator types.

PR #169, opened 2023-09-25
Some Standalone Cleanups

baconpaul:
1. Move to `freeaudio::clap_wrapper::standalone` namespace
2. Remove some trace statements which were distracting when actually using the standalone and which we no longer need.


defiantnerd (approved):
Looks good!

PR #170, opened 2023-09-27
Minor tweaks to support more compiler options

SamWindell:
These changes allow for slightly more flexible compilation settings. Namely: mingw, cross-compilation and C++23.

### DIFF
--- a/include/clapwrapper/vst3.h
+++ b/include/clapwrapper/vst3.h
@@ -25,7 +25,7 @@ typedef uint8_t array_of_16_bytes[16];
 
 
 // VST3GUID allows you to provide the 4 uint32_t parts of the GUID and transforms them to the 16 byte array
-#if WIN32
+#if WIN
 #define VST3GUID(g1, g2, g3, g4) \
 {                                \
 (uint8_t)((g1 & 0x000000FF)      ),    \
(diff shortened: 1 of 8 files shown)


baconpaul:
Thanks

Is there any way we can add these cases to the .github CI actions matrix? I don't know if actions has mingw on win but the matrix is pretty simple. I don't know which compilers have good 23 support but assume gcc-12 does so maybe at least add a gcc12 c++23

I approved you for action runs just now by the way so we can see if this runs with existing set

https://github.com/free-audio/clap-wrapper/blob/18942d7c7e9608f44c0b89006b76f90fb5613e8b/.github/workflows/pullreq.yml#L60

there's the matrix.

baconpaul:
Ahh neat

https://stackoverflow.com/questions/75758572/utilize-pre-installed-mingw-in-github-actions

seems windows-latest has mingw so you can just do a windows matrix entry setting compiler to g++ and gcc; and can also do a gcc12 23 copy also

Do you know how to do that in actions? If you don't I can add it to your PR. Lemme know.

baconpaul (approved):
@defiantnerd these changes are fine. If we want we can just merge this and I can add to the CI matrix for 23 and mingw after that in a separate PR.

Thanks @SamWindell

SamWindell:
Yeah I reckon merge this for now if you're happy with it.

These changes were a result of putting this repo as a git submodule and then using my already existing zig build system to access the VST3 wrapper C++ files and compiling - and it worked very nicely! 

So regarding the CI, I actually don't know if compiling the whole project with mingw/c++23 through cmake needs some tweaks or not. I can perhaps take a look tomorrow if you like.

baconpaul:
Let’s merge this and then I can add a ci stage later tonight to find out. If it works great if it doesn’t then we have a list of things to clean up!

baconpaul:
Also awesome it just works with your zig

defiantnerd:
Thanks @SamWindell !! Much appreciated.

PR #175, opened 2023-10-15
Implement standalone GUI for Windows

mthierman:
This is an initial implementation of a standalone GUI written in win32. Right now we simply handle DPI changes by matching initial dimensions, resizing is implemented but disabled until we can properly check if plugin is resizable.


baconpaul:
I really like how closely you tracked the GTK version structurally. Nice.

I have no idea if the actual win32 code is optimal or not - I'm sure @defiantnerd will have some comments too - and I don't have access to a windows box to try it until Tuesday or so but this is very very exciting. Thank you!

baconpaul:
oh and i just turned on CI so lets see how it does with the suite of compilers! Fun.

baconpaul:
I finally got around to running this and reviewing the code. It's great - simple straightforward starting point.

I recommend we merge it. @defiantnerd - OK with you if we pull this into next so we can keep chipping away at these standalones?

PR #176, opened 2023-10-16
Use clap::helpers::Host

Trinitou:
This is how I think clap::helpers::Host would be used for `clap-wrapper`

If considering merging, please make sure that before
- [x] https://github.com/free-audio/clap-helpers/pull/27 has been merged (contains a build fix when using `clap-helpers`)
- [x] https://github.com/free-audio/clap-helpers/pull/24 has been merged
- [x] include `clap-helpers` using the same pattern as `guarantee_clap` in `base_sdks.cmake`
  -  don't forget to remove the copies!

### DIFF
--- a/cmake/base_sdks.cmake
+++ b/cmake/base_sdks.cmake
@@ -91,6 +91,41 @@ function(guarantee_clap)
     add_subdirectory(${CLAP_SDK_ROOT} base-sdk-clap EXCLUDE_FROM_ALL)
 endfunction(guarantee_clap)
 
+function(guarantee_clap_helpers)
+    if (TARGET clap-helpers)
+        if (NOT TARGET base-sdk-clap-helpers)
+            add_library(base-sdk-clap-helpers ALIAS clap-helpers)
+        endif()
+        return()
+    endif()
+
+
+    if (NOT "${CLAP_HELPERS_SDK_ROOT}" STREQUAL "")
+        # Use the provided root
+    elseif (${CLAP_WRAPPER_DOWNLOAD_DEPENDENCIES})
+        guarantee_cpm()
+        CPMAddPackage(
+                NAME "clap-helpers"
+                GITHUB_REPOSITORY "free-audio/clap-helpers"
+                GIT_TAG "7b53a685e11465154b4ccba3065224dbcbf8a893"
+                EXCLUDE_FROM_ALL TRUE
+                DOWNLOAD_ONLY TRUE
+                SOURCE_DIR cpm/clap-helpers
+        )
+        set(CLAP_HELPERS_SDK_ROOT "${CMAKE_CURRENT_BINARY_DIR}/cpm/clap-helpers")
+    else()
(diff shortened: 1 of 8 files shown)


cmake/base_sdks.cmake
```diff
@@ -91,6 +91,41 @@ function(guarantee_clap)
...
+
+
+    if (NOT "${CLAP_HELPERS_SDK_ROOT}" STREQUAL "")
+        # Use the provided root
+    elseif (${CLAP_WRAPPER_DOWNLOAD_DEPENDENCIES})
+        guarantee_cpm()
+        CPMAddPackage(
+                NAME "clap-helpers"
+                GITHUB_REPOSITORY "free-audio/clap-helpers"
+                GIT_TAG "7b53a685e11465154b4ccba3065224dbcbf8a893"
```
Trinitou, line 110: is this a problem, @baconpaul ?
Actually I'd like to use a version tag / branch here (instead of commit hash) but we do not have that for clap-helpers yet

defiantnerd:
I am not sure if we should include this. The clap-wrapper solves specific problems by "providing the truth" to remove the impedance mismatch between the different plugin standards, like threading models. So checking for those is kind of unnecessary.

Also, it adds another dependency which is something I wanted to avoid: only the SDKs that the wrapper is being built for should be the only ones needed.

Trinitou:
> I am not sure if we should include this. The clap-wrapper solves specific problems by "providing the truth" to remove the impedance mismatch between the different plugin standards, like threading models. So checking for those is kind of unnecessary.

If we would set CheckingLevel to None, those would be gone basically gone. What would remain then? Just the basic C++ wrapper class around clap_host + all clap_host_<extension>s. I think it still would be beneficial even without any checks as the clap-helpers would be the place where new extensions will be added in the future. Then overriding the methods in clap-wrapper will be very fast to do + we can reuse this kind of C-to-C++ wrapping work instead of redoing it for every project again. Also I think it's good to have the clap-wrapper as a kind of standard/example user of clap-helpers who brings in certain requirements/goals for clap-helpers to the table (being zero-overhead, easy to include, etc.).
So I think both clap-wrapper and clap-helpers could benefit from it. But maybe that's not enough?

> Also, it adds another dependency which is something I wanted to avoid: only the SDKs that the wrapper is being built for should be the only ones needed.

Yeah the dependency part is the one I'm also most sceptical about myself. At least @baconpaul has done a great job of leveling-up the clap-helpers CMakeLists.txt so it should now be much more straightforward to have the helpers always in sync with the desired clap-version. (the only missing piece of the puzzle there is the version tag/branch on clap-helpers which already wrote about above)

> I am more in favor to follow the path for the injector plugin to check CLAP conformity between wrapper and target plugin.

As already written, I'd be OK with setting CheckingLevel to None. The question is whether the benefits are enough to convince you to add another dependency. Feel free to do whatever you think is the best of course. In the end it's just a proposal.

Trinitou:
Now I set the `CheckingLevel` to `None`

Trinitou:
@defiantnerd please have look into #177 -> there I now condensed the full idea of using `clap-helpers` in `clap-wrapper`

PR #177, opened 2023-10-16
Use clap::helpers::PluginProxy

Trinitou:
This is how clap::helpers::PluginProxy would be used for `clap-wrapper`

I had to remove the line where the gui extension pointer is set to nullptr if the gui api isn't supported and I'm not sure on how to resolve this.

If considering merging, please make sure that before
- [x] https://github.com/free-audio/clap-helpers/pull/27 has been merged
- [x] CMake files are adjusted to include clap-helpers as well
- [x] Potential conflicts with #176 are resolved

### DIFF
--- a/.github/workflows/pullreq.yml
+++ b/.github/workflows/pullreq.yml
@@ -30,6 +30,7 @@ jobs:
           mkdir deps
           cd deps
           git clone https://github.com/free-audio/clap
+          git clone https://github.com/free-audio/clap-helpers
           git clone https://github.com/steinbergmedia/vst3sdk
           cd vst3sdk
           # temporary workaround, switch to vst3 sdk 3.7.7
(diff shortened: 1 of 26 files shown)


Trinitou:
Ok, now this is almost the full idea of what i want to do, @baconpaul and @defiantnerd :

- deriving Clap::Plugin from `clap;;helpers::Host` (resolves #176 )
- using `clap::helpers::PluginProxy` as convenient C++ handler for the `clap_plugin` the extensions it provides
- using`clap::helpers::MisbehaviourHandler::Ignore` and `clap::helpers::CheckingLevel::None` so there should not be any boilerplate or calls to std::terminate in the final builds (which was something you were concerned about)

The last missing part IMO are the Plugin extension proxies (which are already merged into `clap::helpers`'s `next` Branch but I still get some issues when trying to use them here. Those could be passed instead of `clap::helpers::PluginProxy` in the following places:
- `clap::helpers::PluginAudioPortsProxy` -> `setupAudioBusses`
- `clap::helpers::PluginNotePortsProxy` -> `setupMIDIBusses`
- `clap::helpers::PluginParamsProxy` -> `setupParameters`
- `clap::helpers::PluginGuiProxy` -> `WrappedView`

But I think for now it would work without the extension proxies so I marked this as "ready for review"

If you are still concerned about the additional dependencies, feel free to drop the idea completely. Working on it already brought some improvements for the clap::helpers which is overall a good outcome I think.

PR #186, opened 2023-10-23
Initial DPI scale for win standalone

mthierman:


mthierman:
This is now working, we just need to figure out how to scale a non-resizable plugin.

src/detail/standalone/windows/winutils.cpp
```diff
@@ -238,28 +194,35 @@ void Win32Gui::run()
...
     uint32_t h;
     ui->get_size(p, &w, &h);
 
     Window window;
+    auto dpi{::GetDpiForWindow(window.m_hwnd)};
+    auto scaleFactor{static_cast<float>(dpi) / static_cast<float>(USER_DEFAULT_SCREEN_DPI)};
+
+    if (ui->can_resize(p))
+    {
+      ui->set_scale(p, scaleFactor);
```
baconpaul, line 208: i think we have to call set scale whether you can resize or not, no?

mthierman: Correct. Only the delay (which is not-resizable) seems to not have the correct size now. But it's the same thing in Reaper and Bitwig so I think this matches DAW behavior now.

PR #194, opened 2023-10-31
In explicit (single plugin via cmake config) mode, include UI

baconpaul:
do this by still making a name etc... but skipping the if in the explicit single-plugin mode.


baconpaul:
@defiantnerd we should merge this I think but I realize I didn’t. Any reason to not?

PR #208, opened 2023-12-06
Fix null dereference

SamWindell:
Some hosts call setFrame with a null pointer, for example JUCE-based hosts: https://github.com/juce-framework/JUCE/blob/d054f0d14dcac387aebda44ce5d792b5e7a625b3/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp#L1540C15-L1540C15


baconpaul:
Oooh thanks

PR #211, opened 2024-01-23
updating to CLAP 1.2.0 - removing retired extension midimapping

defiantnerd:
Updates CPM to CLAP 1.2.0, removing an removed draft extension (midimapping)

### DIFF
--- a/cmake/base_sdks.cmake
+++ b/cmake/base_sdks.cmake
@@ -72,7 +72,7 @@ function(guarantee_clap)
         CPMAddPackage(
                 NAME "clap"
                 GITHUB_REPOSITORY "free-audio/clap"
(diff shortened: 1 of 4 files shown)


defiantnerd:
to 1:
the wrapper does not use any extension yet that needs COMPAT. All extensions are stable since ages. I just removed the midimapping which was in there for experiments.

to 2:
They actually don't need to compile with CLAP 1.2.0, it also works with earlier versions. I am not sure if we need it.
Maybe later, but still, the wrapper is from the CLAP HOST perspective. Maybe in the future we need it?

baconpaul:
Great

the compat api ifdef and draft paths are what require a version. If, say, we implement context menu we can add version check

PR #212, opened 2024-01-27
Check for CLAP_PARAM_IS_ENUM

defiantnerd:
implements https://github.com/free-audio/clap-wrapper/issues/89

If a CLAP parameter has CLAP_PARAM_IS_ENUM the VST3 parameter flag kIsList will be added, also the parameter range is applied to the stepping.

### DIFF
--- a/cmake/base_sdks.cmake
+++ b/cmake/base_sdks.cmake
@@ -72,7 +72,7 @@ function(guarantee_clap)
         CPMAddPackage(
                 NAME "clap"
                 GITHUB_REPOSITORY "free-audio/clap"
-                GIT_TAG "1.1.8"
+                GIT_TAG "1.2.0"
                 EXCLUDE_FROM_ALL TRUE
                 DOWNLOAD_ONLY TRUE
                 SOURCE_DIR cpm/clap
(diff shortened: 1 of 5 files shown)


baconpaul:
Did you want this PR to target next rather than main?

[9ab332e](https://github.com/free-audio/clap-wrapper/pull/212/commits/9ab332ea75575b5db34aeb35c47a134ef4a4d27f) looks good but I think it means we require CLAP 1.2.0 where the flag was introduced yes? Shall we do a static assert if you try to build with older clap somewhere more obvious if that's so?

defiantnerd:
Yes, it was intended to be merged to 'next'
The version check macros only apply to. make, right? We don't have ghem in C/C++?

baconpaul:
You can do them in C++

Put this some place "early"

```cpp
#if CLAP_VERSION_LT(1,2,0)
static_assert(false, "CLAP juce wrapper requires at least clap 1.2.0");
#endif
```

`clap.h` defines `CLAP_VERSION_LT`

defiantnerd:
I could confine the usage of CLAP_PARAM_IS_ENUM to if it is `NOT CLAP_VERSION_LT(1,2,0)`.. OTOH 1.2.0 is released now and could just be required (there is no use in compiling against older versions). What do you think?

baconpaul:
I think what we should do is *either* restrict it *or* add a static assert like I showed

the case I want to avoid is someone tries to compile with clap 1.1.8 and gets an unclear error message, which is why I put the check and assert and clear text message in the juce wrappers. 

I think wrapper being on 1.2.0 or greater and having a version check with static assert is best myself. Somewhere early in the compile chain too.

baconpaul:
Looks great!

PR #213, opened 2024-01-31
fsutil: use fs::path for loading libraries to fix win32 unicode paths

SamWindell:
Previously we were using the ASCII versions of some win32 functions. These will cause problems if a filesystem path contains non-English characters.

It made sense to change the Library::load() argument from a const char * to a fs::path for 2 reasons: it ensures paths are kept in the native format (utf16-LE for windows, utf8 for linux/mac), and it avoids the need to do conversions from fs::path to const char * since we were already using fs::path.

### DIFF
--- a/src/detail/clap/fsutil.cpp
+++ b/src/detail/clap/fsutil.cpp
@@ -12,7 +12,7 @@
 #include <cassert>
 #if WIN
 #include <windows.h>
-#include <ShlObj.h>
+#include <shlobj.h>
 #include <sstream>
 #endif
 
@@ -45,13 +45,13 @@ fs::path get_known_folder(const KNOWNFOLDERID &id)
     return {};
 }
 
-std::vector<std::string> split_clap_path(const std::string &in)
+std::vector<std::wstring> split_clap_path(const std::wstring &in)
 {
-  std::vector<std::string> res;
-  std::stringstream ss(in);
-  std::string item;
+  std::vector<std::wstring> res;
+  std::wstringstream ss(in);
+  std::wstring item;
 
-  while (std::getline(ss, item, ';')) res.push_back(item);
+  while (std::getline(ss, item, L';')) res.push_back(item);
 
   return res;
 }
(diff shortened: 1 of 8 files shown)


SamWindell:
Looks like there's a couple of compile errors found in the CI, I'll fix these when I've next got a minute.

baconpaul:
Obviously we need to fix whatever tripped up ci but I think this is a great change. Thank you!

baconpaul:
Oh and once this is sorted probably need to do similar to clap info. That’s more a note to myself than a request for this pr though!

SamWindell:
I fixed the compile error in the CI. It was just a incorrect cast type. So should be good to merge if you're happy with it.

Thanks for your work on this project as well! I do greatly appreciate being able to focus on things other than plugin APIs

baconpaul:
LGTM. @defiantnerd you happy?

baconpaul:
And thanks @SamWindell !

src/detail/clap/fsutil.cpp
```diff
@@ -279,16 +281,19 @@ static void ffeomwe()
 Library::Library()
 {
 #if WIN
-  static TCHAR modulename[2048];
+  fs::path modulename;
   HMODULE selfmodule;
-  if (GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, (LPCTSTR)ffeomwe, &selfmodule))
+  if (GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, (LPCWSTR)ffeomwe, &selfmodule))
```
defiantnerd, line 286: This is not really a string, `GetModuleHandleEx` with the flag `GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS` needs to the address of the function, but you have to fake it to a compatible argument.

defiantnerd:
Does this also work with Mac OS 10.13 and the gulrak::filesystem replacement?

src/detail/clap/fsutil.cpp
defiantnerd: But anyway, this should still work.

defiantnerd:
It looks good for me, thanks @SamWindell

src/detail/clap/fsutil.cpp
SamWindell: Oh yes I see what you mean, I guess it would be fine ascii or wide. No harm either way though

baconpaul:
> Does this also work with Mac OS 10.13 and the gulrak::filesystem replacement?

Ci has that as a config so yes, in general, that’s something the pipeline tests. 

The trick is use fs:: rather than std::filesystem everywhere and then the magic just happens behind the scene

PR #219, opened 2024-02-09
Adding AUv2 Plugin MIDI OUT

defiantnerd:
First implementation of MIDI OUT.

### DIFF
--- a/src/detail/auv2/process.cpp
+++ b/src/detail/auv2/process.cpp
@@ -511,7 +511,6 @@ void ProcessAdapter::addMIDIEvent(UInt32 inStatus, UInt32 inData1, UInt32 inData
       this->_events.emplace_back(n);
       removeFromActiveNotes(&n.note);
       this->output_events_try_push(&this->_out_events, &n.header);
-      _midiouts->send(n);
       break;
     case 9:  // note on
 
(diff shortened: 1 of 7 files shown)


src/detail/auv2/process.cpp
```diff
@@ -540,7 +510,8 @@ void ProcessAdapter::addMIDIEvent(UInt32 inStatus, UInt32 inData1, UInt32 inData
       this->_eventindices.emplace_back((this->_events.size()));
       this->_events.emplace_back(n);
       removeFromActiveNotes(&n.note);
-
+      this->output_events_try_push(&this->_out_events, &n.header);
+      _midiouts->send(n);
```
baconpaul, line 514: This is the only bit I don't understand. Why are we adding to the output events? Is that what `_midiouts` consumes? Also wonder if we want a check around this thing to see if we have midi outs at all? So like in our constructor add a `_hasMidiOuts == !_midiOuts.empty()` and check that here in a branch to avoid the send at all

But I'm just not sure why we added *both* a `try_push` and `send` from reading the code. 

If you can help me understand that rest looks great and we should merge!

defiantnerd: You are correct - this is wrong and I fixed it.

PR #222, opened 2024-02-10
0.7.0

defiantnerd:
* pushed to CLAP 1.2.0
* completed AUv2 wrapper
  * MIDI Out
  * Correct timing
  * Parameter Clumps
* improved standalone wrapper
* improvements regarding sizability and resize request on macOS
* fsutil: use fs::path for loading libraries to fix win32 unicode paths
* fixed host names for wrapper flavors
* removed dependency to removed CLAP_EXT_MIDI_MAPPINGS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,8 @@
 #
 #	clap-wrapper project
 #
-#   Copyright (c) 2022 Timo Kaluza (defiantnerd)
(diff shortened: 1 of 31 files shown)


baconpaul:
This looks correct but please lers rebase not squash and merge it into main. That’s an option in the button above. Then when that’s done reset next to main and move version to 0.8

exciting!

PR #224, opened 2024-02-18
fix: IMIDIMapping wasn't used correctly in VST3 wrapper

defiantnerd:
the init order was wrong, now the wrapper specifics get determined before setting up audio and MIDI busses and parameters.
Since it is a hot fix, the version is being bumped to 0.7.1 and PR goes directly to main.


defiantnerd:
@u-he-frank please have a look if you agree with this.

u-he-frank:
Looks good.

PR #230, opened 2024-02-21
Check active state before querying latency. Fixes #229

### LINKED ISSUE #229: CLAP wrapped VST3 crashing most hosts since CLAP 1.2.0 update
Tested mainly on M1 Sonoma. Reaper also tested on Win 10, same there.
Since the moment Alex updated our u-he plugin builds to CLAP 1.2.0, the wrapped VST3 version will instantly crash most hosts.

The actual CLAP plugin works fine.
The real VST3 works fine.
Only the wrapped VST3 crashes.

It works ok in Bitwig 5.1.2 and Live 11.3.2.

Cubase 13 and Reaper 7 crash as soon as the wrapped VST3 is being loaded.
Studio One might load it, but crashes when quitting the host.
And FL Studio doesn't successfully scan it anymore, so it cannot be loaded there.

The plugin always logs this at the moment of the crash:
`[CLAP] It is wrong to query the latency before the plugin is activated, because if the plugin dosen't know the sample rate, it can't know the number of samples of latency.`

Since most released u-he plugins are not using CLAP 1.2.0 yet, there is only one plugin you can currently test this with, the Zebralette 3 public beta on KVR (revision 15573), which can be downloaded here:

https://www.kvraudio.com/forum/viewtopic.php?t=607153

Schroedingers-Cat:
The VST3 wrapper simply forwards the host's latency request without checking whether the plugin is active. This causes a misbehaving host log when using the CLAP helpers (and is not allowed by CLAP's specs). 

This PR fixes that problem by checking the plugin's active state before getting its latency. This PR also fixes #229.

### DIFF
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -59,6 +59,10 @@ tresult PLUGIN_API ClapAsVst3::terminate()
   if (_plugin)
   {
     _os_attached.off();  // ensure we are detached
+    if (_active)
+    {
+      _plugin->deactivate();
+    }
     _plugin->terminate();
     _plugin.reset();
   }
@@ -111,6 +115,11 @@ tresult PLUGIN_API ClapAsVst3::setActive(TBool state)
 
 tresult PLUGIN_API ClapAsVst3::process(Vst::ProcessData& data)
 {
+  if (!_active || !_processing)
+  {
+    return kNotInitialized;
+  }
+
   auto thisFn = _plugin->AlwaysAudioThread();
   this->_processAdapter->process(data);
   return kResultOk;
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -773,7 +773,9 @@ void ClapAsVst3::request_callback()
 
 void ClapAsVst3::restartPlugin()
 {
-  if (componentHandler) componentHandler->restartComponent(Vst::RestartFlags::kReloadComponent);
+  if (componentHandler)
+    componentHandler->restartComponent(Vst::RestartFlags::kIoChanged |
+                                       Vst::RestartFlags::kLatencyChanged);
 }
 
 void ClapAsVst3::onBeginEdit(clap_id id)
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -87,6 +87,11 @@ tresult PLUGIN_API ClapAsVst3::setActive(TBool state)
         _expressionmap & clap_supported_note_expressions::AS_VST3_NOTE_EXPRESSION_TUNING);
     updateAudioBusses();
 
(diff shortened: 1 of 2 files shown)


baconpaul:
This looks good to me but I think we should merge it into next then cherry pick it to main. But @defiantnerd i am unsure how next v main is tracking right now.

defiantnerd:
I need to think about it - it avoids the protocol violation, but might lead to strange behavior with a wrapped host if it does not ask again for the latency.

Schroedingers-Cat:
> but might lead to strange behavior with a wrapped host if it does not ask again for the latency.  

Even without this fix, the acquired latency will be of questionable correctness since the samplerate isn't known before the plugin is active.  

I can update this PR next Monday with a behaviour that stores a missed latency call and handles that in the next activation of the plugin. That should cover the scenario of a plugin not reporting the latency on its own.

baconpaul:
definitely if you ask an inactive plugin for latency you could queue a latency change update after the active

have we considered just activating the plugin here though also? what's the call path by which this happens?

defiantnerd:
would this also be useful for the tail-extension?

```c++
 auto tailsize = this->_plugin->_ext._tail->get(_plugin->_plugin);
 ```

Schroedingers-Cat:
> have we considered just activating the plugin here though also? what's the call path by which this happens?  

It's a direct call from the host (Reaper in my case, but Cubase 13 is also reported to instantly crash).  

I've updated the PR so that a missed latency-call is stored and acted upon during setActive(). The changes hit an issue with the format checks but no error message is being printed. Any idea why?

baconpaul:
This looks good to me. @defiantnerd shall we include this in 0.7.2 also along with your other changes? If so I can do the double merge and move and stuff.

Schroedingers-Cat:
I've added three additional commits that were still causing crashes during the initialization phase (Reaper, Nuendo). I have no idea how I didn't have these issues last week on my Intel Mac, but these problems were happening on both my Win system and ARM Mac as well as on @defiantnerd 's Win system.  

I'd recommend merging this PR instead of going forward with the incomplete merge via 06e49cfebb540680e18b3c850bdb9c64dc75719b as it's missing another misbehavior fix (bbe9f8994df83a0ac4ef77ae51a56e657aeb9eba) as well as the requested change to store a missed latency call (45b69bc01f77c848c745f0808df322a6d2a64858).

defiantnerd:
it seems that the vst3 documentation contradicts itself and getLatency() is allowed between setupProcessing and setActive - I heard that at least Reaper and Nuendo are doing it that way. So we need to solve this differently by introducing a feature where the CLAP is being transitioned between the necessary states.

defiantnerd:
I'll merge this now. We'll give it a makeover later.

PR #233, opened 2024-02-26
0.7.3

defiantnerd:
- restartPlugin does not lead to kReloadComponent anymore, suits better the semantics of restart_plugin
- ensure param->flush() being called from main_thread
- more documentation on vst3 specific extensions


baconpaul:
Did you mean to close this?

PR #239, opened 2024-03-23
MacOS Standalone Audio Settings Screen; Restartable Audio

baconpaul:
This commit adds an audio settings screen for the macOS standalone, but also updates the generic standalone host to allow audio resets and way more accuracy in setting and managing things like sample rates and devices.

The default behavior is still to bind to system default, and this commit doesn't change that, but does make it editable.

To complete this work some things to do include

- Block size selection and error handling
- Save the state of your setup after you edit it in the patch and settings file
- Add MIDI device selection just like we have audio device selection

But this is an excellent checkpoint into next, so pushing

### DIFF
--- a/src/detail/standalone/entry.cpp
+++ b/src/detail/standalone/entry.cpp
@@ -70,12 +70,6 @@ std::shared_ptr<Clap::Plugin> mainCreatePlugin(const clap_plugin_entry *ee, cons
     }
   }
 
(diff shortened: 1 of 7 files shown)


baconpaul:
Screenshot 2024-03-23 at 12 42 00 PM

Here's the screen by the way.

PR #242, opened 2024-04-19
Adding support for context menus.

defiantnerd:
* adding host side context menu functions
* extending the calling structure/lifecycle management of plugview


baconpaul:
@defiantnerd i reviewed the code and see nothing I’d change. That lifetime mismatch is indeed painful but I think you have it correct. Once the formatting is fixed I Jin his should merge to next

defiantnerd:
Thanks, @baconpaul

PR #245, opened 2024-04-24
Allow child NSView to resize the wrapper NSView in Audio Units.

directmusic:
I was running into an issue where my child NSView would change size and the wrapper would not update its size. This patch addresses this by checking the resize hints provided by the CLAP plugin and setting the AutoresizingMask for the wrapper NSView.

I am quite new to all of this, but my understanding is that AUv2 cannot have a resizable window. Depending on the opinion of the maintainers if you'd prefer it to be consistent with the VST3 and CLAP versions I would be happy to spend the time to implement a listener for mouse events to hittest the edges of the wrapper view to simulate this behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ option(CLAP_WRAPPER_WINDOWS_SINGLE_FILE "Build a single fine (rather than folder
 
 project(clap-wrapper
 	LANGUAGES C CXX
-	VERSION 0.8.0
+	VERSION 0.9.0
 	DESCRIPTION "CLAP-as-X wrappers"
 )
 set(CLAP_WRAPPER_VERSION "${CMAKE_PROJECT_VERSION}" CACHE STRING "Version of the wrapper project")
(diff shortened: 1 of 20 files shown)


baconpaul:
Thanks. @defiantnerd and I are both a bit tied up but I’ll give it a test probably Friday and check

as to a mouse hit test I think if we did that we would want it in an off by default option since we kinda promise to not screw with the events of the underlying clap but I’m interested how you would go about it in the on case?

thank you for the contribution also!

defiantnerd:
Your code looks fine, I'll try to check it out live on macOS tomorrow, too.
Thanks for your contribution.

baconpaul:
@defiantnerd did this get merged? or did deleting next blow up the pr?

defiantnerd:
@baconpaul the rebase blew the PR obviously :/

defiantnerd:
hm, this does not compute.

baconpaul:
i can fix it up

defiantnerd:
I've transfered it to a new PR. Thanks @directmusic

PR #249, opened 2024-04-26
Allow child NSView to resize the wrapper NSView in Audio Units.

defiantnerd:
original contribution by @directmusic


baconpaul:
Yup looks right.

When you commit you can add a co-authored-by with the original author and the log will retain it.

Issue #119, opened 2023-09-09
Include a small simple clap to build and wrap; run pluginval on the result


baconpaul:
1. Take a small simple clap. I'll probably use https://github.com/baconpaul/clap-c99-distortion
2. Make it a subdirectory (but not a sub-linked cmake) of clap-wrapper called "tests/test-clap" and make the cmake do the wrap and so on
3. Build it in CI
4. Have CI download https://github.com/Tracktion/pluginval for the OS in question and run it as a validation step on the resulting VST3, AUV2, etc

basically have CI test that we can build and wrap a micro clap

PR #256, opened 2024-06-10
Add an small example and test of clap-first in-linked wrapper

baconpaul:
- Bring over my c99 distortion and make it C++
- Show that making all flavors as statically linked
- add it to CI

### DIFF
--- a/.github/workflows/pullreq.yml
+++ b/.github/workflows/pullreq.yml
@@ -21,7 +21,7 @@ jobs:
             name: Linux External SDK
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           submodules: recursive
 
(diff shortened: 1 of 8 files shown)


baconpaul:
@defiantnerd this is me looking at the compile flags which tripped up matt and at the same time adding a simple 'clap first' test and example with static linking into the compionents. It's a draft because I want to add some doc and also make CI work. I'll tag you when its ready for review.

baconpaul:
Link to #119 also

baconpaul:
This also shows #257 and #258

baconpaul:
OK @defiantnerd this is now ready for review

Basically what I did was add a small clap and show it linking statically into each of the wrappers, fixed the wrapper over-exports error problem, added some doc, added it to CI

Doing so showed up two errors in the wrapper which i didn't fix in this commit, but will happily pluck at this week.

Let me know what you think! I'm happy to merge this to next if you are.

defiantnerd:
The stepped-bug is not a bug. My guess is that there is a bug in bitwig, but we have to check.

defiantnerd (approved):
Ok by me, so far I get it. I would merge it and we have a final checkup before merging to main.

PR #259, opened 2024-06-10
Fix a VST3 off-by-one error in a stepped param

### LINKED ISSUE #258: VST3 and stepped params
The little simple example I just added has a stepped param with 3 values

The CLAP works correctly in bitwig. The VST3 adds an 'extra' step. That is the mode knob has 4 not 3 values

So we have an off-by-one somewhere.

baconpaul:
The step count is not +1, but rather just the range, as demonstrated by the clap-first distortion plugin compared in bitwig.

Closes #258


baconpaul:
that error is just the win clang screwage

defiantnerd:
Then this is rather a bug in bitwig. I added the +1 when I was debugging stepped parameters in Cubase. So my guess is that bitwig does it wrong. The number of steps of [0..2] is 3, not 2.

steps = (max-min) +1;

PR #261, opened 2024-06-11
Fixing VST3 off-by-one bug in stepped parameters

### LINKED ISSUE #258: VST3 and stepped params

defiantnerd:
The step count is not +1, but rather just the range, as demonstrated by the clap-first distortion plugin compared in bitwig.

Closes #258

### DIFF
--- a/.github/workflows/pullreq.yml
+++ b/.github/workflows/pullreq.yml
@@ -21,7 +21,7 @@ jobs:
             name: Linux External SDK
     steps:
       - name: Checkout code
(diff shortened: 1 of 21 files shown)


baconpaul:
OK I merged 78e46ed into #258 and will merge from there to keep that ARA commit out of the wya. Closing this.

PR #265, opened 2024-07-04
FIx _plugin->_ext._gui can be null on GUI-less plugins

djowel:
auval crashes with segmentation fault on GUI-less AU plugins.

### DIFF
--- a/src/wrapasauv2.cpp
+++ b/src/wrapasauv2.cpp
@@ -676,7 +676,7 @@ OSStatus WrapAsAUV2::GetProperty(AudioUnitPropertyID inID, AudioUnitScope inScop
       case kAudioUnitProperty_CocoaUI:
         LOGINFO("[clap-wrapper] Property: kAudioUnitProperty_CocoaUI {}",
                 (_plugin) ? "plugin" : "no plugin");
-        if (_plugin &&
+        if (_plugin && _plugin->_ext._gui &&
             (_plugin->_ext._gui->is_api_supported(_plugin->_plugin, CLAP_WINDOW_API_COCOA, false)))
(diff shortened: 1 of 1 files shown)


baconpaul:
Thanks!

But I think this is the same change I pushed to next a couple of weeks ago (along with a couple of other protections). Your change is the third one here right? https://github.com/free-audio/clap-wrapper/commit/dbe33d8a14e7567ffe24ca38307855acc3ded5df

baconpaul:
@defiantnerd perhaps we should do a next -> main sweep sooner rather than later to pick up that change if it is the same (and merge this and then do said if this is a case I missed)

djowel:
> Thanks!
> 
> But I think this is the same change I pushed to next a couple of weeks ago (along with a couple of other protections). Your change is the third one here right? [dbe33d8](https://github.com/free-audio/clap-wrapper/commit/dbe33d8a14e7567ffe24ca38307855acc3ded5df)

Right. I wasn't aware of `next`. Should I be working on that branch instead of `main`?

baconpaul:
Ideally we would sweep next to main with some regularity at a release point. We are trying to keep main stable and next as the dev branch

in this case I was just 2 weeks ahead of you in finding the bug 

which you use really depends on your desire to have almost ready commits in your code path

defiantnerd:
We should close this PR and bring `next` to `main`, since it already contains this fix, also on two other places..

baconpaul:
I just synced `next` and `main` and tagged `main` as `v0.10.0`.

As well as ARA support and other fixes, that means main now includes the expanded version of this fix.

Thanks!

Issue #267, opened 2024-07-05
Don't statically link MSVC runtime


mtytel:
https://github.com/free-audio/clap-wrapper/blob/497f40a7580a2f6326fa06691f39aa3d7e0e78a7/CMakeLists.txt#L29

Should just take the default or whatever the plugin developer set it to.

defiantnerd:
We can make this optional - if the wrapper project is the top level project, the library can be statically linked, otherwise just don't?

baconpaul:
yeah that's what we need to do. i just need to figure out the best way to do that cmakewise. i'll get it this weekend.

baconpaul:
Addressed in [48861ec](https://github.com/free-audio/clap-wrapper/commit/48861ecf472fe2cbcd67478e3f5f0dabe9e5785c)

PR #268, opened 2024-07-05
Don't set MSVC to static runtime.

mtytel:
https://github.com/free-audio/clap-wrapper/issues/267

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,6 @@ cmake_minimum_required(VERSION 3.21)
 cmake_policy(SET CMP0091 NEW)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 10.13 CACHE STRING "Minimum macOS version")
 set(CMAKE_OSX_ARCHITECTURES "x86_64;arm64" CACHE STRING "Build Universal Always")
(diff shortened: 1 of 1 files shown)


baconpaul:
So in this case, I think we should set it if it is not otherwise set don't you?

That is, if it is set to nothing set it to static. 

I agree overriding the plugin author is aggravating but people use this as a top level cmake also. Can we if-test it ahead of time?

mtytel:
If you include other cmake directories before clap-wrapper I think you'll get a conflict then and force the plugin dev to set it to DLL or static lib in their project anyway. 
Maybe if this was a full plugin framework it should set it but as a library that's probably included with other libraries I feel like it should just take the default.

baconpaul:
Right. So the problem is we have two use cases

Users like you and me use it as a subordinate library to our project
Users like u-he use it as a standalone to build a wrapper configured to dynaload their remote clap

That's why leaving it unset seems like a tough path for users in the second case. See what I mean?

The top level cmake here does do stuff with command lines basically.

Lemme google a bit and ponder.

baconpaul:
I dealt with this conditionally in [48861ec](https://github.com/free-audio/clap-wrapper/commit/48861ecf472fe2cbcd67478e3f5f0dabe9e5785c) so am going to close this PR.

PR #272, opened 2024-07-07
Version 0.10.0

defiantnerd:
Having ARA complete for VST3, I'd like to name this a 0.10.0 release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@ option(CLAP_WRAPPER_BUILD_TESTS "Build test CLAP wrappers" OFF)
 
 project(clap-wrapper
 	LANGUAGES C CXX
-	VERSION 0.9.2
+	VERSION 0.10.0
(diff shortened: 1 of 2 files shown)


baconpaul:
I tagged main as 0.10.0 last week after we chatted in discord - the tag and changelog are all there and up to date. Did I miss a commit? We can make a 0.10.1 if so

baconpaul:
Ahh yeah I tagged it but forgot to update the cmake :(

baconpaul:
So I think what we should do is

Make this 0.10.1
Merge it
Then tag it 

since I would like tags to be immutable 

agree?

defiantnerd:
I agree.

baconpaul:
Ok I can handle that tonight or tomorrow morning

baconpaul:
OK 0.10.1 is now main with this change, next and main are the same, the tag is there, CMakeLists has the right version, the changeling is updated, and we also included the two changes @mtytel made (VST3 timer and MSVC Runtime optional).

Issue #271, opened 2024-07-06
Support AU 'aumf'


djowel:
Running `auval` gives this warning:

```
Test MIDI
WARNING: AU implements MusicDeviceMIDIEvent but is of type 'aufx' (it should be 'aumf')
  PASS
```

I can't just use 'aumf' because it is not supported and produces build errors.

baconpaul:
Right now if you don’t implement the wrapper factory then we auto detect here 

https://github.com/free-audio/clap-wrapper/blob/497f40a7580a2f6326fa06691f39aa3d7e0e78a7/src/detail/auv2/build-helper/build-helper.cpp#L139

but you can see below thst thst if your clap implements the auv2 wrapper factory you can pick your type directly. 

https://github.com/free-audio/clap-wrapper/blob/497f40a7580a2f6326fa06691f39aa3d7e0e78a7/include/clapwrapper/auv2.h#L9

Alternately if you are a single plugin clap (that is you factory count is 1) you can set the plugin type in cmake. https://github.com/free-audio/clap-wrapper/blob/497f40a7580a2f6326fa06691f39aa3d7e0e78a7/cmake/wrap_auv2.cmake#L12

If these don’t work please do share a bit more detail. Thanks!

djowel:
Pardon the delay. I'll try these suggestions and get back to you on the results. The last one is what I tried (setting the INSTRUMENT_TYPE in cmake).  I get this build error:

```
In file included from /Users/joel/dev/cycfi/soniq/external/clap-wrapper/src/wrapasauv2.cpp:1:
/Users/joel/dev/cycfi/soniq/cmake-build-debug/soniq_auv2-build-helper-output/generated_entrypoints.hxx:6:47: error: expected expression
         free_audio::auv2_wrapper::WrapAsAUV2(,"Soniq", "", 0, ci) {}};
                                              ^
```

The generated_entrypoints.hxx produces an empty string when the type is not handled:

```
#pragma once
#include "detail/auv2/auv2_base_classes.h"

struct wrapAsAUV2_inst0 : free_audio::auv2_wrapper::WrapAsAUV2 {
   wrapAsAUV2_inst0(AudioComponentInstance ci) :
         free_audio::auv2_wrapper::WrapAsAUV2(,"Soniq", "", 0, ci) {}};
AUSDK_COMPONENT_ENTRY(ausdk::AUMusicDeviceFactory, wrapAsAUV2_inst0);
```

[...]

PR #275, opened 2024-07-09
Deal with user specified INSTRUMENT_TYPE

baconpaul:
We switch on INSTRUMENT_TYPE to set the internall wrapper handler. We actually only use this in one spot internally but we were failing to make a buildable wrapper if you set the type to something not in our list-of-three. So for now, warn and default out to internal instrument for the other types (which is probably the right behavior for aumf, which is what raised this issue in #271)


defiantnerd:
yeah, thumbs up. it is build time, so that is a valid choice.

djowel:
Thank you!

PR #276, opened 2024-07-29
Include AUv2 instructions

dathinaios:
Just including the simple instruction cause I had to dig into `CMakeLists.txt` to find out what was needed.


baconpaul:
Thank you and sorry to have missed this. Moved the PR to point at next  and merged

PR #280, opened 2024-08-11
Windows SA: Total refactor

mthierman:
A complete refactor of the Windows standalone

- Now using the W versions of the Win32 API instead of A and have disabled UTF8 code page in the manifest, so we don't need to rely on a minimum version of Windows
- Users can optionally provide a .ico that will display in the titlebar
- Vendored the [Windows Implementation Library](https://github.com/microsoft/wil) to make win32 programming easier (for instance, file dialogs)
- GUI logic should be more robust, still needs some work
- Fixed a bug with audio engine crashing (set the audio in/out counts)
- More correct handling of CL/Clang compiler frontends for entry point
- Fixed CLAP output_name in the standalone wrapper CMake


baconpaul:
So I'm not a windows expert but this looks reasonable to me. Would love @defiantnerd to look

The only thing I'm unsure of is do we want a copy of wil or do we want to CPM it like we do with RTAudio etc...

It is header only so we don't need to compile it in any way, but only grabbing if someone builds windows standalone may be the more symmetric thing to do. Thoughts? May make it more friendly for people being explicit about licenses for sub-components too? Perhaps?

We can also push it to next and change our mind later.

Great stuff!

mthierman:
WIL is now using our guarantee strategy with CPM, and I have fixed some CMake errors and also compiler warnings in /W4.

I've also added more of fmtlib for better logging and improved the GUI code and disabled the WIP settings window.

defiantnerd (approved):
I would take this to next and then test it a bit deeper. On the first glance, all looks very fine..
Thanks @mthierman

PR #285, opened 2024-08-14
Implement SingleComponentEffect::getBusInfo and fix Ardour

baconpaul:
Ardour uses getBusInfo to set up busses even though later we call setBusInfo, and so as a result didn't recieve the kDefaultActive flag we set when we setup processing.

This meant that Ardour would not activate auxilliary busses in multi-out effects so our process loop would find an invalid configuration and not call the clap.

Defacto this meant "no sound in shortcircuit xt in the wrapper in ardour". This imsplementation fixes the issue.


defiantnerd (approved):
Superb! Straight forward.

PR #293, opened 2024-08-17
request_callback impl on windows standalone

MeijisIrlnd:
Spins a timer on construction of the windows standalone, to handle calls to `request_callback`. Timer interval is every 8ms (~= 120hz), kills timer in `onDestroy`.

### DIFF
--- a/src/detail/standalone/windows/helpers.cpp
+++ b/src/detail/standalone/windows/helpers.cpp
@@ -70,6 +70,16 @@ double getScale(::HWND window)
   return static_cast<double>(::GetDpiForWindow(window)) / static_cast<double>(USER_DEFAULT_SCREEN_DPI);
 }
 
+void startTimer(::HWND window, UINT_PTR timerId, UINT intervalMs)
+{
+  ::SetTimer(window, timerId, intervalMs, (::TIMERPROC)NULL);
+}
+
+void stopTimer(::HWND window, UINT_PTR timerId)
+{
+  ::KillTimer(window, timerId);
+}
+
 void abort(unsigned int exitCode)
 {
   ::ExitProcess(exitCode);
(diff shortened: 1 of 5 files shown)


src/detail/standalone/windows/helpers.cpp
```diff
@@ -70,6 +70,16 @@ double getScale(::HWND window)
...
 }
 
+void startTimer(::HWND window, UINT_PTR timerId, UINT intervalMs)
+{
+  ::SetTimer(window, timerId, intervalMs, (::TIMERPROC)NULL);
+}
+
+void stopTimer(::HWND window, UINT_PTR timerId)
+{
+  ::KillTimer(window, timerId);
```
baconpaul, line 80: If set timer fails it returns 0. In that case we probably want to log the failure and not kill timer right?

src/detail/standalone/windows/host_window.cpp
```diff
@@ -300,11 +308,25 @@ int HostWindow::onSysCommand(::HWND hWnd, ::UINT uMsg, ::WPARAM wParam, ::LPARAM
...
   helpers::quit();
 
   return 0;
 }
+
+int HostWindow::onTimerEvent(::WPARAM wParam)
+{
+  if (wParam != s_timerId) return -1;  // or assert fail, etc
+  auto* standaloneHost = freeaudio::clap_wrapper::standalone::getStandaloneHost();
+  const auto hasCallbacks = standaloneHost->callbackRequested.load();
```
baconpaul, line 324: It probably doesn’t matter but std atomic has an atomic compare and ser which would avoid the micro race here of request callback being called again between your load and ser false

baconpaul: So in this case you would do has callbacks o std atomic exchange (&…, false) and not set it to false in the block

baconpaul:
Looks great to me.

PR #294, opened 2024-08-18
macOS standalone request_callback timer

MeijisIrlnd:
Does much the same as the windows PR, but with objc timers - objc isn't my strong suit, so any sanity checking appreciated. 

Also, my previous PR for the `applicationShouldTerminateAfterLastWindowClosed` implementation inadvertently led to a segfault on close, when we reached the section of `applicationWillTerminate` where the window stuff is freed. According to the [docs](https://developer.apple.com/documentation/appkit/nswindow/1419662-close), `[window close]` by default calls `[window release]`, so my working theory is that it was leading to a double free when closed that way. 
To fix it, I implemented `windowWillClose`, and set the window member to `nil` there. Then in `applicationWillTerminate`, I check if `self.window=nil`, and if it *isn't*, call `[[self window] release]`, and that seems to have fixed it!

### DIFF
--- a/src/detail/standalone/macos/AppDelegate.h
+++ b/src/detail/standalone/macos/AppDelegate.h
@@ -6,7 +6,9 @@
 {
   // AudioSettingsWindowDelegate *audioSettingsWindowDelegate;
 }
-@property(assign) IBOutlet NSWindow *window;
+
+@property(assign) NSTimer* requestCallbackTimer;
+@property(assign) IBOutlet NSWindow* window;
 
 - (IBAction)openAudioSettingsWindow:(id)sender;
 
--- a/src/detail/standalone/macos/AppDelegate.mm
+++ b/src/detail/standalone/macos/AppDelegate.mm
@@ -27,6 +27,16 @@
 
 @implementation AppDelegate
 
+- (void)timerCallback:(NSTimer *)instance
+{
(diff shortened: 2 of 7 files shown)


baconpaul:
The objective c code looks correct to me. The lifetime thing I would have to debug but the code looks ok if it stops your crashing i'm sure its good

only thing is this commit also has the windows commit in it ([145c87a](https://github.com/free-audio/clap-wrapper/pull/294/commits/145c87afda22d6433ef808c6cf3bfa55fd5e7fff)) - is that intentional? I thought I had merged that to next so I think it's a rebasing problem more than anything else.  This PR looks like it should just be [b416128](https://github.com/free-audio/clap-wrapper/pull/294/commits/b41612879c1eef2d3640746994bad55b823ed4b2)

MeijisIrlnd:
yeah, I think I must have done something horrible when trying to sync my fork with remote after the last PR got merged, only noticed after I opened this PR - let me figure out what to do about that in the AM!

baconpaul:
here's what you could do.

in your branch (for now) do

```
git checkout next
git fetch upstream
git reset upstream/next --hard
git checkout -b macstuff
git cherry-pick b416128
git push origin mastiff
```

presuming you have upstream set up to point at this repo and origin to point at your fork

PR #296, opened 2024-08-19
CPP 20 flags

baconpaul:
Sigh. CPP20 has .u8string and .string not compatible. Apply the flag in a few places we were missing to fix this with the appropriate generator expression


baconpaul:
Actually hold off I might want to do this differently

baconpaul:
OK f17f5f0 is more like it

Issue #300, opened 2024-08-31
CLAP_PARAM_IS_STEPPED parameter is not working properly


lewloiwc:
When I make this (https://github.com/baconpaul/clap-c99-distortion) into a VST3 using the method described in Getting Started, parameters with CLAP_PARAM_IS_STEPPED don't work properly.
Reaper → https://youtu.be/EfBnUljIXgg
Bitwig → https://youtu.be/cVVahhtgy9o
Live → https://youtu.be/S-7ci_UcmuE

In the CLAP I'm creating, parameters with CLAP_PARAM_IS_STEPPED also don't work properly, so I think it might be an issue with clap-wrapper. Can you reproduce this in your environment?

I'm using Windows 10, and for building, I'm using:
Developer Command Prompt for VS 2022
Visual Studio 2022 Developer Command Prompt v17.10.5

defiantnerd:
You are right. Thanks for bringing this up.

You can fix this by changing line 121 in src/detail/vst3/parameter.cpp from:
`auto param_range = (info->max_value - info->min_value);`
to 
`auto param_range = (info->max_value - info->min_value) + 1`;

I will fix this asap in the `next` branch.

defiantnerd:
it seems that it is an issue with the value string, the parameter itself is working well (you can hear the three different settings).
I'll check it out if I can reproduce that.

defiantnerd:
I am able to reproduce it and this is the fix:
```diff
--- a/src/detail/vst3/parameter.h
+++ b/src/detail/vst3/parameter.h
@@ -54,7 +54,7 @@ class Vst3Parameter : public Steinberg::Vst::Parameter
   {
     if (info.stepCount > 0)
     {
-      return (vst3value * info.stepCount + 1) + min_value;
+      return (vst3value * info.stepCount) + min_value;
     }
     return (vst3value * (max_value - min_value)) + min_value;
   }
```
I'll prepare a fix for the wrapper, too.

PR #301, opened 2024-09-01
off by one error in value conversion

defiantnerd:
@baconpaul this is a fix for issue https://github.com/free-audio/clap-wrapper/issues/300

I wonder if the function below `inline double asVst3Value(double clapvalue) const` the `+1` needs to be removed, too.

### DIFF
--- a/src/detail/vst3/parameter.h
+++ b/src/detail/vst3/parameter.h
@@ -50,6 +50,10 @@ class Vst3Parameter : public Steinberg::Vst::Parameter
   bool fromString(const Steinberg::Vst::TChar* string, Steinberg::Vst::ParamValue& valueNormalized) const override;
 #endif
 
+  // for asClapValue() and asVst3Value()
+  // regarding conversion and the meaning of stepCount take a look here:
+  // https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Parameters+Automation/Index.html#conversion-of-normalized-values
+
   inline double asClapValue(double vst3value) const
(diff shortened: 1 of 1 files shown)


baconpaul:
So step count "1" means to positions one gap.
So the function below for a 0..1 value would give a VST3 value of 0...0.5 because of the +1
So I agree with you that other +1 has to go

defiantnerd:
ok, checked out the documentation here:
https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Parameters+Automation/Index.html#conversion-of-normalized-values

Discrete Value => Normalize
`double normalized = discreteValue / (double)stepCount;`

Normalize => Discrete Value (Denormalize)
`int discreteValue = min (stepCount, normalized *(stepCount + 1));`

I'll take a look if that is properly converted here.

baconpaul:
Doesn't that mean Discrete(Normalize(val)) != val? since it is discrete/stepcount * (stepcount+1)

lewloiwc:
I was also curious about the Discrete(Normalize(val)) != val issue, so I tried to organize it in a diagram. (I apologize for it being extremely difficult to read.)
image
At first, I also thought that Discrete(Normalize(val)) != val would be true, but it seems that the rounding that occurs due to "int discreteValue = x" was important.
Therefore, I think this error can probably be fixed by removing the +1 from both asClapValue and asVst3Value.
When I actually created a VST3 with the version where I removed these, it seems to be working correctly so far.

PR #306, opened 2024-09-07
Remove CFRelease() calls in host_get_name()

### LINKED ISSUE #304: AUv2: Crash when quickly loading multiple copies on M1/M3 Mac
Hello, I have been doing some stress testing for the plugins I am working on and noticed that in Live 12.0.1 (as well as 12.1b2) on my M1 MacBook after around the 4th instance of the AUv2 being inserted Live would crash. Initially, it appeared like it was crashing in my GUI, so I disabled "Auto-Open Plug-In Windows" in Live's settings. When attaching Xcode to live it crashes on this line:

https://github.com/free-audio/clap-wrapper/blob/e82e4d7628fe962a3df4bbecff0120bee7d39c31/src/detail/auv2/auv2_base_classes.h#L456

Here is the partial stacktrace:
```
MainThread (1) Queue : com.apple.main-thread (serial)
#0	0x000000018206c14c in CF_IS_OBJC ()
#1	0x0000000181f23dd8 in CFStringGetCString ()
#2	0x000000034cfe47e8 in free_audio::auv2_wrapper::WrapAsAUV2::host_get_name() at /Users/joe/Documents/GitHub/MiniMeters/build/_deps/clap-wrapper-src/src/detail/auv2/auv2_base_classes.h:456
#3	0x000000034dafce3c in Clap::Plugin::Plugin(Clap::IHost*) at /Users/joe/Documents/GitHub/MiniMeters/build/_deps/clap-wrapper-src/src/clap_proxy.cpp:175
#4	0x000000034dafc92c in Clap::Plugin::Plugin(Clap::IHost*) at /Users/joe/Documents/GitHub/MiniMeters/build/_deps/clap-wrapper-src/src/clap_proxy.cpp:184
#5	0x000000034dafc81c in Clap::Plugin::createInstance(clap_plugin_factory const*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, Clap::IHost*) at /Users/joe/Documents/GitHub/MiniMeters/build/_deps/clap-wrapper-src/src/clap_proxy.cpp:142
#6	0x000000034cfde410 in free_audio::auv2_wrapper::WrapAsAUV2::WrapAsAUV2(AUV2_Type, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, int, ComponentInstanceRecord*) at /Users/joe/Documents/GitHub/MiniMeters/build/_deps/clap-wrapper-src/src/wrapasauv2.cpp:135
```

Additionally this shows up:
```
Application Specific Information:
CF objects must have a non-zero isa
```
When googling that string I found this thread that may explain that this occurs on macOS 14 due to the fact they "added extra hardening to Core Foundation in macOS 14": https://forums.developer.apple.com/forums/thread/739355

I revisited the crash in my GUI code and - while I definitely have a bug there - the stacktrace similarly ends with CFStringGetCString() while macOS tries to generate the crash report inside "com.apple.metal.telemetry".

I can reproduce this in Reaper, Hosting AU, FL Studio 2024 but I have not been able to reproduce this in Logic.
Returning `_hostname.c_str()` at the beginning of the `get_host_name()` stops the crashes.

I can also reproduce this crash using the ClapFirstDistortion Test plugin and it crashes on the same line, or in the call at auv2_base_classes.h:437.

**Reproduction steps:**
* Set Live to not "Auto-Open Plug-In Windows" -- for consistency's sake.
* Build the test plugin and move to the correct folder
* Add an instance in Live
* Duplicate instance as quickly as possible.

It seems like this may be an issue with Audio Units as a whole, but I am not sure. I was unable to make my JUCE plugin crash the same way. However I noticed some other discussions:
https://discussions.apple.com/thread/254500745?sortBy=rank
https://community.native-instruments.com/discussion/19674/logic-crashes-when-loading-certain-libraries-in-kontakt-7

So far my tests included:
* M1 Max MacBook Pro - macOS 14.6.1 - Crash
* M3 Pro MacBook Pro (a friend's machine) - macOS 14.6.1 - Crash

The commands I used to build the project.
```
cmake .. -DCLAP_WRAPPER_DOWNLOAD_DEPENDENCIES=1 -DCLAP_WRAPPER_BUILD_TESTS=1
make
```

**Possible workarounds:**

1. One thing that did work was making all of the `CFStringRef` and the `CFBundleRef` static inside the `get_host_name()` method and not calling `CFRelease()` on any of them. I never had a crash the first load. There must be some edge case inside of CoreFoundation where calling `CFBundleGetValueForInfoDictionaryKey()` fails silently where the isa field is null, but the pointer returned from myVersionString (and similar vars) is not null.

2. Getting the `_hostname` string once when creating the factory. As stated above I was not able to make it crash on the first instance.

Thank you for your time.

Edit: Just tested this in a macOS 13 VM on my machine and it also crashes the same way.

directmusic:
In the Core Foundation functions any time the word "Get" is used we do not have ownership and do not need to call CFRelease(). Fixes #304

https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-SW1

### DIFF
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -426,17 +426,19 @@ class WrapAsAUV2 : public ausdk::AUBase,
   const char* host_get_name() override
   {
     char text[65];
+    // No need to release any of these "Get" functions.
+    // https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-SW1
 
     CFBundleRef applicationBundle = CFBundleGetMainBundle();
     if (applicationBundle != NULL)
     {
       CFStringRef myProductString =
           (CFStringRef)CFBundleGetValueForInfoDictionaryKey(applicationBundle, kCFBundleNameKey);
+
       if (myProductString)
       {
         CFStringGetCString(myProductString, text, 64, kCFStringEncodingUTF8);
         _hostname = text;
-        CFRelease(myProductString);
       }
       else
       {
@@ -445,10 +447,8 @@ class WrapAsAUV2 : public ausdk::AUBase,
         {
           CFStringGetCString(applicationBundleID, text, 64, kCFStringEncodingUTF8);
           _hostname = text;
-          CFRelease(applicationBundleID);
         }
       }
-      //  CFRelease(applicationBundle);  Don't release it
       CFStringRef myVersionString =
           (CFStringRef)CFBundleGetValueForInfoDictionaryKey(applicationBundle, kCFBundleVersionKey);
       if (myVersionString)
@@ -457,12 +457,12 @@ class WrapAsAUV2 : public ausdk::AUBase,
         _hostname.append(" (");
         _hostname.append(text);
         _hostname.append(")");
-        CFRelease(myVersionString);
       }
-      _hostname.append(" (CLAP-as-AUv2)");
+      _hostname.append(" (CLAP-as-AUv2-wrapper)");
     }
     return _hostname.c_str();
   }
+
 
   // --------------- IAutomation
   void onBeginEdit(clap_id id) override;
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -458,11 +458,10 @@ class WrapAsAUV2 : public ausdk::AUBase,
         _hostname.append(text);
         _hostname.append(")");
       }
-      _hostname.append(" (CLAP-as-AUv2-wrapper)");
+      _hostname.append(" (CLAP-as-AUv2)");
     }
     return _hostname.c_str();
   }
-
 
   // --------------- IAutomation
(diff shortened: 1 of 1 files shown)


baconpaul:
formatting aside (and that it came from main not next so blew out a tiny formatting change on next) this change looks correct to me. I may tweak it a touch tomorrow to fix formatting and remove the string change I had made on the hostname. but @defiantnerd wdyt?

And thanks @directmusic  - we should review the rest of the AUV2 code to make sure we aren't making a similar error elsewhere.

baconpaul:
Actually had a moment before end of day and pushed up a clang format fix and also the regression of next vs main

if CI passes lets ask @defiantnerd to look.

Thank you!!

defiantnerd:
Yeah, fix is fine. Thank you very much, @directmusic !
And yes, we should give AUv2 another review for this issue.

Issue #188, opened 2023-10-24
Standalone Features: Windows


mthierman:
Development ocurring on my fork @ https://github.com/mthierman/clap-wrapper

# Must haves
- [x] Settings (Audio/MIDI Ins & Outs)
- [x] Standalone state
- [x] Plugin state
  - [x] Save/Load defaults
  - [x] Save/Load patches
  - [x] Reset
- [x] MsgBox for errors
- [x] App manifest
- [x] Convert from console app to win32 app
- [x] Debug console
- [x] Hide console earlier
- [x] Constrain window if plugin is not resizable
- [x] ~~Fullscreen~~
- [x] ~~Dark mode~~
- [x] PowerShell script to convert .PNG to .ICO
  - [x] CMake script to create .rc file (for .ico) and add to sources
    - https://github.com/defiantnerd/neui/blob/main/neui/cmake/resources.cmake
    - https://bolu.dev/programming/2020/07/18/Cmake-icons.html

# Nice to haves
- [ ] ~~Support multi-plugins(?)~~
- [ ] ~~Support plugins without a GUI~~
- [ ] ~~About CLAP plugin window~~

# Bugs to fix
- [x] Fix crashes from some .clap files (U-He, AudioThing)
- [x] ~~clap-saw-demo crash on exit because of VSTGUI::exit()~~ [clap-saw-demo bug..]
- [x] `clap_plugin_gui.can_resize() was called without a prior call to clap_plugin_gui.create()` in WM_WINDOWPOSCHANGED

PR #307, opened 2024-09-19
Windows SA: Audio/MIDI settings

mthierman:
Apologies for the semi-crazy diff.

1. Windows standalone is now a single header/source unit, and the main() file has been renamed and cleaned up.
 - Implemented a message manager to make construction and handling messages easier
 - Audio/MIDI settings now work and state is saved as a JSON file in the same location as the plugin autosave and default states. We can also save plugin window position now
 - GUI handling has been much improved and most of the bugs have been fixed. There may be some remaining bugs in regards to resizing, however.
 - We are using Windows.Data.Json (using C++/WinRT that ships with the Windows SDK to avoid another dependency) for the settings

I have tried to be as minimally invasive to the standalone host to implement changing audio ins/outs, buffer sizes & sample rates, and MIDI devices. This should be double checked against the mac and linux versions for any breakage but the behavior should be the same on those platforms.


mthierman:
Closes #188.

src/detail/standalone/entry.cpp
```diff
@@ -75,6 +76,7 @@ std::shared_ptr<Clap::Plugin> mainCreatePlugin(const clap_plugin_entry *ee, cons
       if (fs::exists(loadPath / "settings.clapwrapper"))
       {
         LOG << "Trying to load from clap wrapper settings" << std::endl;
+        fs::create_directories(loadPath);
```
baconpaul, line 79: This creat isn't needed since we already know the directory exists from the if above

mthierman: Good catch, I forgot we were loading here.. Removed it.

Issue #305, opened 2024-09-07
Windows+VST3 shows garpled parameter names and values with special characters.


lewloiwc:
The "°" character in "180°" and similar expressions becomes garbled.
However, this might be my own issue and may not be relevant to others. I'm not sure where exactly the problem lies.

I know how to solve it in my environment, but I lack knowledge about character encoding related to differences in OS and compilers. There's probably an issue somewhere, and I'm not sure if this is the correct approach, but this resolved the character garbling for me.
I confirmed that the garbled characters were fixed on Windows 10 & (Ableton Live 12 | Bitwig Studio 5.2.3 | Cubase 13 | FL Studio 2024 | REAPER 7.22 | Studio One 6).

These are using the rewritten version → https://chatgpt.com/share/cb8268e7-cb34-48f4-9783-76e8b6e6d89a

Fixing garbled parameter names
---
https://github.com/free-audio/clap-wrapper/blob/48861ecf472fe2cbcd67478e3f5f0dabe9e5785c/src/detail/vst3/parameter.cpp#L95
Change this to:
```cpp
{
  const char *utf8_text = fullname.c_str();
  uint32_t out_buffer_capacity = 128;
  uint32_t out_index = 0;
  for (uint32_t i = 0;out_index < out_buffer_capacity - 1 && utf8_text[i] != 0;) {
    uint32_t codepoint = 0;
    uint8_t byte = static_cast<uint8_t>(utf8_text[i]);
    
    if ((byte & 0b10000000) == 0b00000000) {
      codepoint = byte;
      i += 1;
    } else if ((byte & 0b11100000) == 0b11000000) {
      codepoint = byte & 0b00011111;
      codepoint = (codepoint << 6) | (static_cast<uint8_t>(utf8_text[i + 1]) & 0b00111111);
      i += 2;
    } else if ((byte & 0b11110000) == 0b11100000) {
      codepoint = byte & 0b00001111;
      codepoint = (codepoint << 6) | (static_cast<uint8_t>(utf8_text[i + 1]) & 0b00111111);
      codepoint = (codepoint << 6) | (static_cast<uint8_t>(utf8_text[i + 2]) & 0b00111111);
      i += 3;
    } else if ((byte & 0b11111000) == 0b11110000) {
      codepoint = byte & 0b00000111;
      codepoint = (codepoint << 6) | (static_cast<uint8_t>(utf8_text[i + 1]) & 0b00111111);
[...]

PR #312, opened 2024-09-25
replacing insufficient VST3 SDK conversion with proper version.

defiantnerd:
see issue https://github.com/free-audio/clap-wrapper/issues/305

### DIFF
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -42,6 +42,84 @@ struct ClapHostExtensions
 };
 #endif
 
+void utf8_to_utf16l(const char* utf8string, uint16_t* target, size_t targetsize)
+{
+  bool result = true;
+  uint32_t codepoint = 0;
+  int state = 1;
+  size_t targetpos = 0;
+
+  auto src = reinterpret_cast<const uint8_t*>(utf8string);
+  size_t pos = 0;
+  while (src[pos] && (targetpos < (targetsize - 1)))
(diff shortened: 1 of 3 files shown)


src/wrapasvst3.cpp
```diff
@@ -42,6 +42,82 @@ struct ClapHostExtensions
...
 #endif
 
+void utf8_to_utf16l(const char* utf8string, uint16_t* target, size_t targetsize)
+{
+  uint32_t codepoint = 0;
+  size_t targetpos = 0;
+
+  auto src = reinterpret_cast<const uint8_t*>(utf8string);
+  size_t pos = 0;
+  while (src[pos] && (targetpos < (targetsize - 1)))
```
defiantnerd, line 52: `(targetsize -2)` because a character and NUL must have room.

baconpaul (approved):
This all looks fine to me. The only question is are we sure `Steinberg::char16` and `uint16_t` are always the same type.

I would be tempted to add somewhere

```
static_assert(std::is_same_v<Steinberg::char16, uint16_t>);
```

just in case an API shifts underneath us. But glad to hear why we don't need that too!

lewloiwc:
The parameter names were displayed correctly when I tried building with the code here!! Please continue with the parameter values as well! Once again, thank you all for working on this problem!

defiantnerd:
Just checked in that values and bus names are also correctly converted.

PR #316, opened 2024-10-10
Add CLI to wrap CLAP plugins

CrushedPixel:
~~DRAFT PR. DO NOT MERGE. DISCUSSION REQUIRED.~~
Any feedback is welcome.

This still has several flaws:
- largely untested on Windows
- ~~Standalone doesn't work on macOS either~~ Standalone is unsupported atm as we can't build it into single files on all platforms.
- ~~uses CPM to get fmt when we already vendor it, cause otherwise the program doesn't link~~
- ~~all the metadata going into AU needs to be removed once it's read from the CLAP. This should happen before this gets merged~~

Needs to be discussed:
- ~~replace OS APIs with std::filesystem at the expense of bumping min macOS version to 10.15?~~


CrushedPixel:
The error when building standalone:

```
[ 87%] Building OBJCXX object CMakeFiles/gain_standalone.dir/_deps/clap-wrapper-src/src/detail/standalone/macos/AppDelegate.mm.o
/Users/marius/Documents/CLAPPlugins/clap-wrapper/cmake-build-debug/cli/wrapped/wrap-build/_deps/clap-wrapper-src/src/detail/standalone/macos/AppDelegate.mm:41:24: error: use of undeclared identifier 'HOSTED_CLAP_NAME'
   41 |   std::string clapName{HOSTED_CLAP_NAME};
      |                        ^
```

Is it even possible to wrap a CLAP plugin into a single dependency-free standalone binary on Windows? If not, do we even want to support this from the CLI?

cli/src/main.cpp
```diff
@@ -0,0 +1,267 @@
...
+        COMMAND ${CMAKE_COMMAND} -E copy
+            $<TARGET_FILE:${STANDALONE_TARGET}>
+            "${OUTPUT_DIR}/${PROJECT_NAME}"
+    )
+  endif()
+endif()
+)";
+
+  cmake_file << cmake_content;
+  cmake_file.close();
```
CrushedPixel, line 177: probably not needed, would happen implicitly in the destructor

cli/src/main.cpp
```diff
@@ -0,0 +1,267 @@
...
+  {
+    throw std::runtime_error("Failed to create CMakeLists.txt");
+  }
+
+  const char* cmake_content = R"(
+cmake_minimum_required(VERSION 3.15)
+project(${PROJECT_NAME})
+
+set(CMAKE_CXX_STANDARD 17)
+if (APPLE)
```
CrushedPixel, line 77: also only add if AUv2 wrapper is enabled

cli/src/main.cpp
```diff
@@ -0,0 +1,267 @@
...
+    // construct cmake command with variables supplied
+    std::string cmake_cmd = fmt::format(
+        "cmake . "
+        "-DPROJECT_NAME=\"{}\" "
+        "-DCLAP_PLUGIN_PATH=\"{}\" "
+        "-DOUTPUT_DIR=\"{}\" "
+        "-DBUILD_AU={} "
+        "-DBUILD_VST3={} "
+        "-DBUILD_STANDALONE={} "
+        "-DAU_SUBTYPE_CODE=\"{}\"",
```
CrushedPixel, line 245: should be removed once this is inferred from the CLAP. @baconpaul

baconpaul:
OK so a couple of things looking

1. We can use `filesystem` just we need to use either ghc or built in. The wrapper does this now and has a target for it.
2. I gotta figure out why fmt didn't work

so let me pull this branch down now and see if i can fix that and push up a diff or two to get you started.

CrushedPixel:
At this point, I believe what's left to do is:
- testing on Windows
- writing a README

Please let me know if there's any more work to do.

cli/src/main.cpp
```diff
@@ -67,11 +67,18 @@ if(BUILD_VST3)
...
+  if (APPLE)
+    add_custom_command(TARGET ${VST3_TARGET} POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_directory
+            $<TARGET_BUNDLE_DIR:${VST3_TARGET}>
+            "${OUTPUT_DIR}/${PROJECT_NAME}.vst3"
+    )
+  else()
+    add_custom_command(TARGET ${VST3_TARGET} POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_directory
+            "${OUTPUT_DIR}/${PROJECT_NAME}.vst3"
```
baconpaul, line 79: This can't work right? IT copies from nowhere to the output dir?

cli/src/main.cpp
```diff
@@ -0,0 +1,332 @@
...
+    {
+      throw std::runtime_error(
+          "AU manufacturer code must be exactly 4 characters long, and start with an uppercase letter "
+          "followed only by lowercase letters and/or numbers");
+    }
+  }
+#else
+  if (build_au)
+  {
+    throw std::runtime_error("AU output format is only supported on macOS");
```
baconpaul, line 203: Is 'throw' the right thing here? This message won't appear on the CLI for most folks. I think we should print and exit(1) or some such

cli/src/main.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+    // write CMakeLists.txt
+    write_cmakelists(build_dir);
+
+    // extract project name from input path
+    auto project_name = sanitize_cmake_identifier(input_path.stem());
+
+    // construct cmake command with variables supplied
+    std::string cmake_cmd = fmt::format(
+        "cmake . "
```
baconpaul, line 235: We probably want an argument for cmake path which defaults to `cmake`

baconpaul (commented):
Some comments on a quick morning skim

cli/src/main.cpp
```diff
@@ -0,0 +1,332 @@
...
+  {
+    create_directory_if_not_exists(output_dir);
+    create_directory_if_not_exists(output_dir);
+    create_directory_if_not_exists(build_dir);
+
+    // write CMakeLists.txt
+    write_cmakelists(build_dir);
+
+    // extract project name from input path
+    auto project_name = sanitize_cmake_identifier(input_path.stem());
```
baconpaul, line 231: So this line is what causes your windows break in CI.

On windows, `std::filesystem::path` does not cast to `std::string` since the native type on windows is a `std::wstring`. 

The idiom you are looking for here is `input_path.stem().u8string()` which will convert to a UTF8 representation of the underlying file system.

cli/src/main.cpp
CrushedPixel: My bad, I coded as if we're in the `try` block. Will fix

cli/src/main.cpp
CrushedPixel: Whoops. Should be fixed now, but requires some testing on Windows, so I won't resolve convo yet

sourcebox:
Testing the CLI with Linux Mint 22 shows two issues when building a VST3: 
- a build error `recompile with -fPIC` that can be fixed with setting `export CXXFLAGS="$CXXFLAGS -fPIC"`
- the generated bundle has a wrong directory structure. The shared library has to be put under `Contents/x86_64-linux` to make it appear in Bitwig Studio.

Issue #287, opened 2024-08-14
Wrapper sets `_host->version` to the plugin version not the host version


baconpaul:
In short circuit I construct my clap about screen this way (in my local branch - merging soon)

```cpp
engine->runningEnvironment =
        "Clap-First " + std::string(h->name) + " " + std::string(h->version);
```

where `h` is my `clap_host_t *` passed in at construction. Fine.

The CLAP in bitwig shows this

![Screenshot 2024-08-14 at 9 47 09 AM](https://github.com/user-attachments/assets/1d110a9a-bf88-4e75-80c6-86e34758e6b2)

The VST3 wrapped in bitwig though shows this

![Screenshot 2024-08-14 at 9 47 25 AM](https://github.com/user-attachments/assets/a7354aea-1643-4db9-aeb0-5522a00d84c1)

(0.7.225.0 is my *plugin* version number)

and the AUv2 wrapped in logic shows the same

![Screenshot 2024-08-14 at 9 48 00 AM](https://github.com/user-attachments/assets/10347963-38db-4059-a178-55ffbf9c63f2)

(once I merge #286 to remove the specious -wrapper string)

so I think we need to go and use the protocols to find the running host version and report it if possible. Or at least report the clap wrapper version rather than the plugin version.

defiantnerd:
Those strings are actually only in debug builds.

baconpaul:
Are you sure?  I think you are confusing the plugin name with the clap `_host->name`

https://github.com/free-audio/clap-wrapper/blob/48861ecf472fe2cbcd67478e3f5f0dabe9e5785c/src/detail/auv2/auv2_base_classes.h#L462

that's not debug guarded

https://github.com/free-audio/clap-wrapper/blob/48861ecf472fe2cbcd67478e3f5f0dabe9e5785c/src/wrapasvst3.cpp#L1083

and neither is that. And I just did a release build and the screen for the VST3 was unchanged

PR #321, opened 2024-10-12
Update the wrapper host display to be more accurate

baconpaul:
This commit does a few things

1. Makes the wrapper host author and URL the free audio locations
2. Sets the wrapper host versio properly to the WRAPPER version not to the PLUGIN version, when used as a subordinate cmake, by using PROJECT_VERSION rather than CMAKE_PROJECT_VERSION to set the ifdef
3. Reads the human readable string out of the bundle for auv2 version which makes the host read as "Logic Pro 11.0.1 (Clap-as-AUV2)" rather than "Logic Pro (6029) (Clap-as-AUv2)".
4. Ups the cmake version to 0.11.0 in anticipation of the move of next to main

Addresses #287


src/clap_proxy.cpp
```diff
@@ -173,8 +173,8 @@ Plugin::Plugin(IHost* host)
   : _host{CLAP_VERSION,
           this,
           host->host_get_name(),
-          "defiant nerd",
```
defiantnerd, line 176: 😸

defiantnerd (approved):
Yeah, all fine.

src/clap_proxy.cpp
baconpaul: Yeah I thought about a list but decided to just teamify it. Hope that’s ok!

PR #323, opened 2024-10-14
Windows standalone: Use default entry point, unicode argv

mthierman:
Switches argv to unicode.

Main function is now the default, wWinMain (which is more correct and allows us to skip setting /ENTRY linker flags, reducing CMake complexity).

### DIFF
--- a/src/detail/standalone/windows/windows_standalone.cpp
+++ b/src/detail/standalone/windows/windows_standalone.cpp
@@ -2,6 +2,22 @@
 
 namespace freeaudio::clap_wrapper::standalone::windows_standalone
 {
+std::pair<int, std::vector<std::string>> getArgs()
+{
+  int argc{0};
+  wil::unique_hlocal_ptr<wchar_t*[]> buffer;
+  buffer.reset(::CommandLineToArgvW(::GetCommandLineW(), &argc));
+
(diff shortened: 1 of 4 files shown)


src/detail/standalone/windows/windows_standalone.cpp
```diff
@@ -2,6 +2,22 @@
 
 namespace freeaudio::clap_wrapper::standalone::windows_standalone
 {
+std::pair<int, std::vector<std::string>> getArgs()
```
messmerd, line 5: Why return a `std::pair`? `std::vector` already stores its size.

mthierman: Indeed, in fact the size is not even really necessary, mainCreatePlugin isn't even using it. It's actually quite unfortunate that I need to construct 2 vectors for this when I could just pass a string. I've updated the PR to remove the std::pair

defiantnerd:
@mthierman there is a conflict left, it looks like there should be both upstreams should be accepted. Do you agree?

mthierman:
Oh man, this has been sitting around for a year so I'm not sure. I'll have a look when I have some time

mthierman:
Closing in favor of #434

Issue #324, opened 2024-10-30
Eject au tags


baconpaul:
https://developer.apple.com/documentation/audiotoolbox/audio_components

PR #325, opened 2024-10-30
Eject the CLAP features as AUv2 tags

baconpaul:
Addresses #324

### DIFF
--- a/src/detail/auv2/build-helper/build-helper.cpp
+++ b/src/detail/auv2/build-helper/build-helper.cpp
@@ -85,8 +85,13 @@ struct auInfo
     {
       of << "        <key>tags</key>\n"
          << "        <array>\n";
-      for (const auto &tag : tags)
+      for (auto tag : tags)  // purposefully take a copy so we can mutate case
       {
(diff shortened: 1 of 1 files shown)


defiantnerd:
Looks fine - just a thought: should we make first character uppercase?

this is all lowercase: https://github.com/free-audio/clap/blob/27f20f81dec40b930d79ef429fd35dcc2d45db5b/include/clap/plugin-features.h#L39

and the Apple examples have first character uppercase.

what do you think?

baconpaul:
Yeah I thought about that but my instinct was to not mess with them. But i think you are correct.

It's super easy to do an `if a < tag[0] < z tag0 = topper tag0` and probably a smart choice.

Let me toss that in.

baconpaul:
OK pushed.

Shortcircuit now gets

```
  <key>tags</key>
        <array>
          <string>Instrument</string>
          <string>Sampler</string>
          <string>Synthesizer</string>
          <string>Free and Open Source</string>
        </array>
      </dict>
```

defiantnerd (approved):

Issue #332, opened 2025-01-05
VST3 and requestedFlush; wrong but double wrong in reaper


baconpaul:
params::requestFlush basically sets _requestFlush

https://github.com/free-audio/clap-wrapper/blob/705df2703fbb991b3609e1a5c91d92a5c2213925/src/wrapasvst3.cpp#L1214

that then calls flush here

So there's two problems

1. I think that should being if _processing and not if _active inside the condition
2. but even with that change Reaper wont work because Reaper has a weird thing when you toggle a vst3 off that it calls setProcessing(false) then immediately setprocessing(true).

This is the cause of https://github.com/baconpaul/six-sines/issues/54

baconpaul:
https://forum.cockos.com/showthread.php?p=2835482#post2835482 asked the reaper devs there

baconpaul:
Schwa and I bang it out on that thread. Basically we need to deal with setProcessing always being true in vst3s while still doing flush. I have a plan.

PR #333, opened 2025-01-05
Partial fix to VST3 flush

baconpaul:
This is a partial fix to the VST3 flush issue we are facint.  #332. 

It is a partial fix for two reasons

1. The use of _processEverCalled doesn't preclude the race if we actually *are* flushing. We need antoerh atomic pair / spin lock Alas because of reaper we need that extra every called check since the prior check on _active was too conservative and a check just on _processing will blow up reaper

2. With this in place, we flush and don't crash, but don't send automation events. Compare the way deactivated six sines works in bitwig with and without this. The CLAP will send automatino events when deactivated and the VST3 wont. This is because of the lack of an _automation and _vstdata structure to generate events.

Maybe we merge this into next and work forward or maybe not but wanted to get this in place for reivew

### DIFF
--- a/src/detail/shared/spinlock.h
+++ b/src/detail/shared/spinlock.h
@@ -1,0 +1,41 @@
+
+#pragma once
+#include <atomic>
+namespace ClapWrapper::detail::shared
+{
+struct SpinLock
+{
+ private:
+  std::atomic<bool> locked_{false};
+
+ public:
+  SpinLock() : locked_(false)
+  {
+  }
+
+  void lock()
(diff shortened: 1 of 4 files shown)


baconpaul:
OK with [3fb8cd5](https://github.com/free-audio/clap-wrapper/pull/333/commits/3fb8cd533674d4a7d2a849d892498f91245b6d92) we are at least thread safe.

defiantnerd (approved):
Looks good for me.

PR #336, opened 2025-01-08
Create portable CIDs on Mac/Lin

baconpaul:
The automatic creation of CIDs using the VST3 SDK produced different GUID orders on MAC/LIN vs WIN (COM_COMPATBILE). Basically if COM_COMPATBILE the VSt3 SDK in absense of an explicit bit pattern onto the FUID would do an 8 bit endian then pair of 4 bit endian flips. This manifested in our vst3.h extension also.

So clean all this up. With this change setting an expilcit CID using CMake or the extension, or using the calculated one from the id with the hash, generates the same CID on mac/win/lin so load compatability across platforms is preserved.

This also means VST3 generated prior to this diff using the automatic mechanism will have their CID changed on macos and linux. Before we merge to Main we may want to allow a control that avoids this skip, or investigate whether a VSt3 can have two CIDs and then advertise both. But we should get this to next asap.

### DIFF
--- a/include/clapwrapper/vst3.h
+++ b/include/clapwrapper/vst3.h
@@ -24,9 +24,19 @@ static const CLAP_CONSTEXPR char CLAP_PLUGIN_AS_VST3[] = "clap.plugin-info-as-vs
 typedef uint8_t array_of_16_bytes[16];
 
 // clang-format off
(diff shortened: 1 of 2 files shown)


baconpaul:
As well as testing those scenarios, with this change in place, I can confirm that a windows reaper project including six sines will load the six sines vst3 on mac with this change applied, and won't with the change not applied.

Issue #326, opened 2024-12-27
VST3+Cubase 14 mono tracks are not working properly


lewloiwc:
Tested with Windows 10, Cubase 14, CLAP 1.2.2, latest clap-wrapper, and clap-c99-distortion.

When using a VST3 on mono tracks in Cubase 14, the `process()` function, which is usually called, is instead replaced by calls to `flush()`, causing the plugin to malfunction.  

As long as the track remains mono, `process()` is never called, and `flush()` is called instead. However, when switching the track to stereo, `process()` is called as expected, and everything works normally.  

In Cubase 13, process() was called correctly even on mono tracks, but crashed at `const float in_r = process->audio_inputs[0].data32[1][i];`. (Can't verify now as Cubase 13 trial expired)

---

Separate question: The current clap-c99-distortion plugin crashes or throws errors when used on mono tracks in Cubase or Studio One. I’m not sure how to prevent these crashes or errors with minimal changes.

For now, I’m using this method, but it’s clear that something is wrong. Based on my research, I focused on the fact that valid pointers seem to fall within the 0xFFFFFFFFF address range:
```diff
@@ clap-c99-distortion.c @@
-	389	const float in_r = process->audio_inputs[0].data32[1][i];
+	389	const float in_r = reinterpret_cast<uintptr_t>(process->audio_inputs[0].data32[1]) <= 0xFFFFFFFFF ? process->audio_inputs[0].data32[1][i] : 0.0f;
-	430	process->audio_outputs[0].data32[1][i] = out_r;
+	430	reinterpret_cast<uintptr_t>(process->audio_outputs[0].data32[1]) <= 0xFFFFFFFFF ? process->audio_outputs[0].data32[1][i] = out_r;
```

When I tried to test this newly created code in Cubase 14, I found this bug. So, I haven’t been able to run this code yet, but I feel like there’s something wrong with this approach as well:
```diff
@@ clap-c99-distortion.c @@
-	389	const float in_r = process->audio_inputs[0].data32[1][i];
+	389	const float in_r = process->audio_inputs_count >= 2 ? process->audio_inputs[0].data32[1][i] : 0.0f;
-	430	process->audio_outputs[0].data32[1][i] = out_r;
+	430	process->audio_inputs_count >= 2 ? process->audio_outputs[0].data32[1][i] = out_r;
```
I think there’s probably something like `process->audio_inputs[0].data32_count` somewhere, but I couldn’t find it.

baconpaul:
[...]

PR #346, opened 2025-01-22
Survive probing with bus layouts

baconpaul:
We know our channel count; if we get asked for a bus layout in the VST3 with a different channel count, we should say no. So when we get a call to setBusArrangement, pull the bus info, construct a speaker arrangement (although probably some ambisonic work to do here in the future) and say no if i get a channel count i don't support

I think this will solve #326
I know it will solve https://forum.cockos.com/showthread.php?p=2840258#post2840258


baconpaul:
Conversation on this pr on discord amounts to

1. it matches the conservatism we have downstream but
2. It’s not optimistic enough about other topologies which
3. We should read from the various extensions and 
4. Allow to pad up (so give a mono to stereo option or pad missng option)
5. 

I’ll write that up in an issue when I merge it later tonight

PR #365, opened 2025-03-19
WCLAP option for clap-first

geraintluff:
Uses Emscripten, and only tested on Mac so far.

If you don't have Emscripten set up:
```sh
git clone https://github.com/emscripten-core/emsdk.git path/to/emsdk
cd path/to/emsdk
./emsdk install latest
./emsdk activate latest
```

Then set the environment variables:
```sh
# set up
. path/to/emsdk/emsdk_env.sh
```

When _creating_ the CMake build, use the `emcmake` helper (which uses the appropriate toolchain):

```sh
emcmake cmake -B build-emscripten
```

Then you build as normal:
```sh
cmake --build build-emscripten --target my-plugin_wclap --config Release
```


baconpaul:
Hi!

I repointed to 'next' and started CI. I'll try and find time to review today (or perhaps @defiantnerd can look also but this looks deep in the cmake, which was more my doing than his).

Thanks!

cmake/make_clapfirst.cmake
```diff
@@ -99,7 +100,7 @@ function(make_clapfirst_plugins)
     set(ALL_TARGET ${C1ST_TARGET_NAME}_all)
     add_custom_target(${ALL_TARGET})
 
-    if (${BUILD_CLAP} GREATER -1)
+    if (NOT EMSCRIPTEN AND ${BUILD_CLAP} GREATER -1)
```
baconpaul, line 103: How is the 'EMSCRIPTEN' variable set?

cmake/wrap_wclap_emscripten.cmake
```diff
@@ -0,0 +1,39 @@
...
+#       set(EMSCRIPTEN 1)
+#   endif()
+
+function(target_add_wclap_configuration)
+    set(oneValueArgs
+            TARGET
+            OUTPUT_NAME
+    )
+    cmake_parse_arguments(TCLP "" "${oneValueArgs}" "" ${ARGN} )
+
```
baconpaul, line 15: Should we have something like

```
if (NOT ENSCRIPTEN)
   message(FATAL_ERROR "Do not call this without emscrpten running")
endif()
```

This is sort of my same as above. Who is responsible for setting that flag which lets us detect if we are running emscripten?

cmake/make_clapfirst.cmake
baconpaul: Also is there a way to structure this so not every BUILD_ requires the NOT below? Like we could wrap the 

```
 list(FIND C1ST_PLUGIN_FORMATS "CLAP" BUILD_CLAP)
     list(FIND C1ST_PLUGIN_FORMATS "VST3" BUILD_VST3)
     list(FIND C1ST_PLUGIN_FORMATS "AUV2" BUILD_AUV2)
```

into 

```
if (EMSPCRIPTE)
   set(BUILD_CLAP -1)
   ...
else()
   list(FIND ...)
endif()
```

that way when we add a format we can remember the enscrimpten if as opposed to not.

baconpaul (changes requested):

baconpaul:
OK CI passed (good!) but I had two questions and one suggestion which I put in as a review.

baconpaul:
Oh also wondering if we want to add a CI stage to test this, but we can do that separately from here since i'm sure it involves some annoying GitHub actions work.

cmake/make_clapfirst.cmake
geraintluff: Ah yes!  That makes sense.

geraintluff: At the moment, I'm setting `EMSCRIPTEN` manually.  However, we could instead check `CMAKE_SYSTEM_NAME` as set in [the Emscripten toolchain](https://github.com/emscripten-core/emscripten/blob/66d2137b0381ac35f7e2346b2d6a90abd0f1211a/cmake/Modules/Platform/Emscripten.cmake#L17).

geraintluff: The code would look like:

```cmake
if (CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
	set(EMSCRIPTEN 1)
endif()
```

@baconpaul Where's the best place for this to go?

geraintluff:
Example plugin build as WCLAP: https://signalsmith-audio.github.io/wasm-clap-browserhost/?module=plugin/clapfirst-bad-distortion.wclap.tar.gz

The other one (Six Sines) wouldn't be as enlightening since I currently don't have note-event input in that browser-based WCLAP host.  One day!

baconpaul:
> Example plugin build as WCLAP: https://signalsmith-audio.github.io/wasm-clap-browserhost/?module=plugin/clapfirst-bad-distortion.wclap.tar.gz
> 
> The other one (Six Sines) wouldn't be as enlightening since I currently don't have note-event input in that browser-based WCLAP host. One day!

This is so awesome!

PR #368, opened 2025-03-20
Clean up PROJECT_IS_TOP_LEVEL defaults and remove unused CMake options

### LINKED ISSUE #278: CMake build fails if we don't set VST3 path or set CLAP_WRAPPER_DOWNLOAD_DEPENDENCIES=TRUE
Just adding the clap-wrapper to a project will fail if we don't provide a VST3 SDK. Unsure if we should be setting CLAP_WRAPPER_DOWNLOAD_DEPENDENCIES by default or, probably, do nothing.

```cmake
[cmake] CMake Error at D:/Repos/free-audio/clap-wrapper/cmake/base_sdks.cmake:42 (message):
[cmake]   Unable to detect vst3sdk! Have you set -DVST3_SDK_ROOT=/path/to/sdk or
[cmake]   CLAP_WRAPPER_DOWNLOAD_DEPENDENCIES=TRUE
[cmake] Call Stack (most recent call first):
[cmake]   D:/Repos/free-audio/clap-wrapper/cmake/base_sdks.cmake:115 (search_for_sdk_source)
[cmake]   D:/Repos/free-audio/clap-wrapper/cmake/wrap_vst3.cmake:53 (guarantee_vst3sdk)
[cmake]   D:/Repos/free-audio/clap-wrapper/CMakeLists.txt:102 (target_add_vst3_wrapper)
[cmake] 
[cmake] 
[cmake] -- clap-wrapper: searching for 'vst3sdk' in "D:/Repos/free-audio/clap-wrapper"...
[cmake] -- Configuring incomplete, errors occurred!
```

mthierman:
Fixes #278


cmake/shared_prologue.cmake
```diff
@@ -86,7 +86,7 @@ endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
     target_compile_options(clap-wrapper-compile-options-public INTERFACE /utf-8 /Zc:__cplusplus)
-    if (${CMAKE_CXX_STANDARD} GREATER_EQUAL 20)
```
baconpaul, line 89: This could just be

```
if ("${CMAKE_CXX_STANDARD}" GREATER_EQUAL 20)
```

and the quote enclosure will deal with non defined

cmake/top_level_default.cmake
```diff
@@ -0,0 +1,52 @@
...
+	if (${CLAP_WRAPPER_BUILD_STANDALONE})
+		add_executable(${pluginname}_as_standalone)
+		target_add_standalone_wrapper(TARGET ${pluginname}_as_standalone
+		    OUTPUT_NAME ${CLAP_WRAPPER_OUTPUT_NAME}
+		    HOSTED_CLAP_NAME ${CLAP_WRAPPER_OUTPUT_NAME}
+		    PLUGIN_INDEX 0)
+	endif()
+endif()
+
+if (${CLAP_WRAPPER_BUILD_TESTS})
```
baconpaul, line 50: We probably want these three lines back in the top level cmake and not buried here. I sometimes turn on tests when running as a lib for submodules.

cmake/shared_prologue.cmake
mthierman: Great. It seems like we assume this is being set globally anyway (supporting target_compile_features will probably be more involved) so it doesn't matter much anyway

PR #374, opened 2025-03-28
Windows filesystem related fixes and minor refactor

SamWindell:
Sorry this just one big set, I'm a bit stretched for time at the moment for teasing this into separate PRs.

I've been using these changes in Floe's alpha for a while.

There's a few things here (the commit comments are pretty descriptive too).

## Fix GetModuleFileNameW 
The way we were getting the filepath of a DLL was just straight up not working (a change I made in a previous PR). We were doing this:
```
    std::wstring p;
    auto size = GetModuleFileNameW(selfmodule, nullptr, 0);
    p.resize(size);
    size = GetModuleFileNameW(selfmodule, &p[0], size);
```

But GetModuleFileNameW doesn't ever return the needed size. So we need a new way of doing things.

## Fix edge cases and minor refactor
There's also a couple of edge cases that are now properly handled and I made usage of fs::path more universal.

Let me know of any changes.

### DIFF
--- a/src/detail/auv2/build-helper/build-helper.cpp
+++ b/src/detail/auv2/build-helper/build-helper.cpp
@@ -5,14 +5,7 @@
 #include <sstream>
 
 #include "detail/clap/fsutil.h"
-
-#if MACOS_USE_STD_FILESYSTEM
-#include <filesystem>
-namespace fs = std::filesystem;
-#else
-#include "ghc/filesystem.hpp"
-namespace fs = ghc::filesystem;
-#endif
+#include "detail/os/fs.h"
 
 struct auInfo
 {
--- a/src/detail/clap/fsutil.cpp
+++ b/src/detail/clap/fsutil.cpp
@@ -305,7 +305,7 @@ Library::Library()
           GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
           (LPCWSTR)ffeomwe, &selfmodule))
   {
-    if (this->getEntryFunction(selfmodule, getModulePath(selfmodule).u8string().c_str()))
+    if (this->getEntryFunction(selfmodule, os::getModulePath(selfmodule).u8string().c_str()))
     {
       _selfcontained = true;
     }
--- a/src/detail/clap/fsutil.cpp
+++ b/src/detail/clap/fsutil.cpp
@@ -305,11 +305,19 @@ Library::Library()
           GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
           (LPCWSTR)ffeomwe, &selfmodule))
   {
-    std::wstring p;
-    auto size = GetModuleFileNameW(selfmodule, nullptr, 0);
-    p.resize(size);
-    size = GetModuleFileNameW(selfmodule, &p[0], size);
-    modulename = std::move(p);
+    DWORD bufferSize = MAX_PATH;
+    std::wstring buffer(bufferSize, L'\0');
+
+    DWORD length = GetModuleFileNameW(selfmodule, buffer.data(), bufferSize);
+
+    while (length == bufferSize && GetLastError() == ERROR_INSUFFICIENT_BUFFER)
+    {
+      bufferSize *= 2;
+      buffer.resize(bufferSize);
+      length = GetModuleFileNameW(selfmodule, buffer.data(), bufferSize);
+    }
+    buffer.resize(length);
+    modulename = std::move(buffer);
   }
   if (selfmodule)
   {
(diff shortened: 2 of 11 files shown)


src/detail/standalone/windows/windows_standalone.cpp
```diff
@@ -1275,7 +1275,7 @@ bool Plugin::saveSettings()
 
   if (settingsPath.has_value())
   {
-    std::filesystem::create_directories(settingsPath.value() / plugin.plugin->desc->id);
+    fs::path::create_directories(settingsPath.value() / plugin.plugin->desc->id);
```
baconpaul, line 1278: I think this change is wrong; shouldn't it be `fs::create_directories` not `fs::path::create_directories`? `create_directories` is a free function in `std::filesystem` namespace.

src/detail/clap/fsutil.cpp
```diff
@@ -293,21 +293,31 @@ Library::Library()
...
-    p.resize(size);
-    size = GetModuleFileNameW(selfmodule, &p[0], size);
-    modulename = std::move(p);
+    DWORD bufferSize = MAX_PATH;
+    std::wstring buffer(bufferSize, L'\0');
+
+    DWORD length = GetModuleFileNameW(selfmodule, buffer.data(), bufferSize);
+
+    while (length == bufferSize && GetLastError() == ERROR_INSUFFICIENT_BUFFER)
+    {
```
baconpaul, line 314: I don't think there's a big risk of runaway here, but do we want to cap length at MAX_PATH or some such in this while? Just being paranoid really. 

That is `while (length < MAX_PATH && ....` or similar?

src/detail/clap/fsutil.cpp
```diff
@@ -293,21 +293,31 @@ Library::Library()
 #if STATICALLY_LINKED_CLAP_ENTRY
   _pluginEntry = &clap_entry;
   _selfcontained = true;
-  std::string path = os::getModulePath();
-  setupPluginsFromPluginEntry(path.c_str());
+  fs::path path = os::getPluginPath();
```
baconpaul, line 296: Just idiomatically I would probably have used auto here. No need to change it just tossing in the comment (and won't repeat it elsewhere)

src/detail/auv2/build-helper/build-helper.cpp
```diff
@@ -5,14 +5,7 @@
...
 #include "detail/clap/fsutil.h"
-
-#if MACOS_USE_STD_FILESYSTEM
-#include <filesystem>
-namespace fs = std::filesystem;
-#else
-#include "ghc/filesystem.hpp"
-namespace fs = ghc::filesystem;
-#endif
+#include "detail/os/fs.h"
```
baconpaul, line 8: This is a wonderful change we should have made forever ago!

baconpaul (commented):
This looks great. I had two questions which may need a change. Also I'm not a windows expert so wouldn't mind @defiantnerd just giving it a once over for that part too.

src/detail/clap/fsutil.cpp
SamWindell: You're right an upper bound would be reassuring. I'll add that. I will pick something larger than MAX_PATH though, because Windows can work beyond that size. https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later

defiantnerd: I have a better idea, let me add that..

defiantnerd: The upper bound is 32k on windows.

defiantnerd:
Ok, lets merge that.

src/detail/clap/fsutil.cpp
baconpaul: Ugh. TIL about https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

baconpaul: oh which you linked also. ugh. "Its MAX_PATH unless its not and then its 32k but maybe a bit less depending, and if you want to know if it is MAX_PATH or not, its not, unless it is, depending" is roughly

SamWindell:
I'll fix the compile error and make the changes suggested. Maybe Tuesday

src/detail/clap/fsutil.cpp
mthierman: > But GetModuleFileNameW doesn't ever return the needed size. So we need a new way of doing things.

I added WIL as a dependency for the standalone, and there is a [helper](https://github.com/microsoft/wil/wiki/Win32-helpers#functions-returning-variable-sized-strings) for this common and annoying Windows API quirk =)

I'm fine with whatever as long as long paths are supported!

mthierman: I'm not sure a cap is necessary.. my understanding is the function will fail and return 0 if there is a problem. The WIL helpers don't have a cap either

SamWindell:
Added the things we talked about.

defiantnerd (approved):
Thanks, Sam, this looks very fine and cleans things up much!

baconpaul:
Agree @defiantnerd  - merging.

PR #380, opened 2025-04-10
`RESOURCE_DIRECTORY` for copying verbatim into clap-first bundle

geraintluff:
This works for MacOS CLAP/VST3.

It *should* print out a warning for all other platforms, saying this option is not supported.

### DIFF
--- a/cmake/make_clapfirst.cmake
+++ b/cmake/make_clapfirst.cmake
@@ -42,6 +42,7 @@ function(make_clapfirst_plugins)
             WINDOWS_FOLDER_VST3 # True for a filder, false for single file (default)
 
             ASSET_OUTPUT_DIRECTORY
+            RESOURCE_DIRECTORY # Entire directory to be copied into a bundle/folder's resources
 
             AUV2_MANUFACTURER_NAME # The AUV2 info. If absent we will probe the CLAP for the
             AUV2_MANUFACTURER_CODE # auv2 extension
@@ -64,6 +65,10 @@ function(make_clapfirst_plugins)
 
     if (NOT DEFINED C1ST_WINDOWS_FOLDER_VST3)
         set(C1ST_WINDOWS_FOLDER_VST3 FALSE)
+    endif()
+
+    if (NOT DEFINED C1ST_RESOURCE_DIRECTORY)
+        set(C1ST_RESOURCE_DIRECTORY "")
     endif()
 
     if (NOT DEFINED C1ST_PLUGIN_FORMATS)
@@ -122,6 +127,7 @@ function(make_clapfirst_plugins)
                 OUTPUT_NAME ${C1ST_OUTPUT_NAME}
                 BUNDLE_IDENTIFIER "${C1ST_BUNDLE_IDENTIFIER}.clap"
                 BUNDLE_VERSION "${C1ST_BUNDLE_VERSION}"
+                RESOURCE_DIRECTORY "${C1ST_RESOURCE_DIRECTORY}"
         )
         if (DEFINED C1ST_ASSET_OUTPUT_DIRECTORY)
             if (NOT WIN32)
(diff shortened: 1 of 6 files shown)


baconpaul:
So this basically assigns a resource directory argument to each sub target. And looks correct.

 But it makes me wonder if instead, since these 93 changes will be irraitatinv for the next 14 properties. If we want to contemplate something like a cmake function which sets properties on all the wrapper generated targets based on the clap first name. Something like `set_Clapfirst_Properties(target t formats f1 f2 properties a avail b Bval)112th

then we could use that after make clap first to markup the targets.

see what I mean?

geraintluff:
I do see how setting properties (on the `_impl` static library?) would mean new settings propagate nicely to all the functions.

I feel that's a separate (and substantial!) refactor though, not for this PR.

baconpaul:
Nan I was thinking of a new function which sets on the endpoints. 
‘let me peek tomorrow and if it is a nightmare we can just merge this. That ok with you?

geraintluff:
Sounds great. 😄 

A note/question: because there are custom copying commands for both `RESOURCE_DIRECTORY` and `COPY_AFTER_BUILD`, there are two (conditional) calls to `add_custom_command()`.  This worked OK when I tested it, but it also smells like a race condition.

baconpaul:
I thought (and it’s my experience so far that it is true that) cmake reliably orders them in parse order so we should be fine but a test in ci would be great

geraintluff:
@baconpaul Hey - did you manage to take a look at this?

baconpaul:
Oh I'm so sorry

I like my bigger idea but won't get to it soon so lets go with my original instinct of 'rebase, pass CI, and merge'

sound good?

geraintluff:
@baconpaul Bump

cmake/make_clapfirst.cmake
```diff
@@ -220,10 +230,9 @@ function(make_clapfirst_plugins)
         endif()
 
 	# Emscripten flags
-        set_target_properties(${WCLAP_TARGET} PROPERTIES
-		COMPILE_FLAGS "-msimd128" # TODO: add this to the static library as well?
-        	LINK_FLAGS    "-msimd128 -sSTANDALONE_WASM --no-entry -s EXPORTED_FUNCTIONS=_clap_entry,_malloc -s INITIAL_MEMORY=512kb -s ALLOW_MEMORY_GROWTH=1 -s ALLOW_TABLE_GROWTH=1 -s PURE_WASI --export-table"
-        )
+	target_compile_options(${C1ST_IMPL_TARGET} PUBLIC -msimd128 --no-entry)
```
baconpaul: Ahh i was just about to press merge when I saw this. Why do we add this to all C1ST implements targets? That seems like an awfully big change.

I would like to have a more general way like I mentioned but let's just merge this so you have the feature and I stop forgetting. But on a review that caught my eye.

baconpaul:
Oh damnit I'm so sorry

I was about to merge it and on a review saw one thing not about RESOURCE_CIRECTORY which I don't think belongs in this commit and looks wrong.

The resource directory only changes I'm happy to merge even though I think there's a more general blah blah way we could do but since we haven't lets get this feature in your hands now.

cmake/make_clapfirst.cmake
geraintluff: Assuming you mean `-msimd128` - it has wide support across modern browsers and WASM engines.  See "Fixed-width SIMD" on [this page](https://webassembly.org/features/).

I can take it out, but it seemed (to me) like a reasonable baseline, and has enough performance impact to be worth adding.  It's roughly comparable to the SSE level of SIMD native SIMD stuff.

geraintluff: It was in there as a TODO (next line), so I must've done it and then got it caught up in this PR by accident, when I was refactoring to avoid deprecated CMake variables (`COMPILE_FLAGS`/`LINK_FLAGS`).

geraintluff:
@baconpaul Thanks for catching that - I've removed the extra commit.

PR #382, opened 2025-04-18
Use CMAKE_SYSTEM_PROCESSOR when building the VST3 directories

directmusic:
This patch simply removes hard-coded "x86_64" when creating the .vst3 directory structure on Windows and Linux and replaces it with the `CMAKE_SYSTEM_PROCESSOR` variable.

### DIFF
--- a/cmake/wrap_vst3.cmake
+++ b/cmake/wrap_vst3.cmake
@@ -142,6 +142,8 @@ function(target_add_vst3_wrapper)
         endif()
 
     elseif(UNIX)
+        message(STATUS "clap-wrapper: Building the VST3 Bundle Folder using the CMAKE_SYSTEM_PROCESSOR variable: (${CMAKE_SYSTEM_PROCESSOR})")
+
         target_link_libraries(${V3_TARGET} PUBLIC "-ldl")
         target_link_libraries(${V3_TARGET} PRIVATE "-Wl,--no-undefined")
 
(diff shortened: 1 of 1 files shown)


baconpaul:
This is a clever and appropriate change 

my only question is reading https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html#variable:CMAKE_SYSTEM_PROCESSOR it’s not clear that the variable is always set to a valid vst3 architecture 

should we throw a warning if the architecture isn’t x86-64 or aarch64 or arm64ec? Continue but warn? 

Or warn in that case just if the system processor is the host processor?

directmusic:
Looking at the documentation [here](https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Locations+Format/Plugin+Format.html) I notice they have some specific platform strings listed for Windows, but for Linux they say it should be `uname -m` result + "-linux".

It seems to me like that checking should be rigorous on Windows, but maybe less so on Linux?

baconpaul:
> Looking at the documentation [here](https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Locations+Format/Plugin+Format.html) I notice they have some specific platform strings listed for Windows, but for Linux they say it should be `uname -m` result + "-linux".
> 
> It seems to me like that checking should be rigorous on Windows, but maybe less so on Linux?

yes that sounds right. especially since from cmake doc it seems like the odd cases are really things like msvc in 32 bit mode on an amd and stuff

so check but perhaps also perhaps put out a `message(STATUS `  even if the check is ok. messages always help when we get bug reports.

baconpaul:
Great!

PR #384, opened 2025-04-19
Silence warnings when parent project has fmt-header-only target

directmusic:
The `fmt-header-only` target in fmt defines `FMT_HEADER_ONLY` as `1` as seen [here](https://github.com/fmtlib/fmt/blob/b8192d233a4015144ba42be201ba15d9af8dfbdb/CMakeLists.txt#L379). However, throughout its use in clap-wrapper it is defined without the "1" leading to some warnings (or in my case errors). 

```
/Users/joe/Documents/GitHub/MiniMeters/external/clap-wrapper/src/detail/clap/../os/log.h:3:9: error: 'FMT_HEADER_ONLY' macro redefined [-Werror,-Wmacro-redefined]
    3 | #define FMT_HEADER_ONLY
      |         ^
<command line>:1:9: note: previous definition is here
    1 | #define FMT_HEADER_ONLY 1
      |         ^
1 error generated.
```

The simple fix is just to change the definitions in log.h and windows_standalone.h which is what this patch does.

### DIFF
--- a/src/detail/os/log.h
+++ b/src/detail/os/log.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#define FMT_HEADER_ONLY
+#define FMT_HEADER_ONLY 1
 #include "fmt/format.h"
 #include "fmt/ranges.h"
 
--- a/src/detail/standalone/windows/windows_standalone.h
+++ b/src/detail/standalone/windows/windows_standalone.h
@@ -21,7 +21,7 @@
 #include <winrt/windows.foundation.h>
 #include <winrt/windows.data.json.h>
 
-#define FMT_HEADER_ONLY
+#define FMT_HEADER_ONLY 1
 #include <fmt/format.h>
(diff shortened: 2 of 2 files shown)


baconpaul:
Ugh I thought I fixed all of these but perhaps I fixed them the wrong way. 

C++ is terrible but I’m happy to follow their cmake. I at least made it consistent in 12.0 but perhaps I picked the wrong side

directmusic:
C++ _is_ terrible. If it wasn't apparent, I've been just doing plumbing on my own builds the last few days so I definitely feel that haha.

If you feel this is not a comprehensive fix, alternatively we could add a definition in the check for the `fmt-header-only` target and guard around the `FMT_HEADER_ONLY` define in those files, but that may be over complicating it.

baconpaul:
I guess these could be `#if !definee FMT_HEADER_ONLY #define` right? Would that work? I think it would?

Issue #383, opened 2025-04-19
Build issue on Clang 20.1.2 due to outdated fmt headers


directmusic:
I am running into build issues when using the the latest Zig compiler - which is backed by Clang 20.1.2 - for cross builds to Windows from macOS. The errors I receive have been addressed in more recent fmt versions. I dropped the headers from fmt v11.1.4 in and that fixes these issues.

Is there any opposition to updating those headers to the latest fmt version (11.1.4)? If not, I would be happy to open a pull request. 

The errors (I removed the full paths for brevity):

```
clap-wrapper/libs/fmt/fmt/format-inl.h:59:24: error: call to consteval function 'fmt::basic_format_string<char, fmt::basic_string_view<char> &, const char (&)[3]>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
clap-wrapper/libs/fmt/fmt/format-inl.h:60:22: error: call to consteval function 'fmt::basic_format_string<char, const char (&)[7], int &>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
clap-wrapper/libs/fmt/fmt/format-inl.h:1387:35: error: call to consteval function 'fmt::basic_format_string<char, unsigned int &>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
clap-wrapper/libs/fmt/fmt/format-inl.h:1391:33: error: call to consteval function 'fmt::basic_format_string<char, unsigned int &>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
clap-wrapper/libs/fmt/fmt/format-inl.h:1394:33: error: call to consteval function 'fmt::basic_format_string<char, int>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
clap-wrapper/libs/fmt/fmt/format-inl.h:59:24: error: call to consteval function 'fmt::basic_format_string<char, fmt::basic_string_view<char> &, const char (&)[3]>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
clap-wrapper/libs/fmt/fmt/format-inl.h:60:22: error: call to consteval function 'fmt::basic_format_string<char, const char (&)[7], int &>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
[...]

PR #385, opened 2025-04-19
Update libfmt to v11.1.4

directmusic:
This pull simply brings in the latest version of libfmt (v11.1.4) into clap-wrapper and addresses the build error I mentioned in the issue I opened here: https://github.com/free-audio/clap-wrapper/issues/383

### DIFF
--- a/cmake/wrap_vst3.cmake
+++ b/cmake/wrap_vst3.cmake
@@ -142,6 +142,8 @@ function(target_add_vst3_wrapper)
         endif()
 
     elseif(UNIX)
+        message(STATUS "clap-wrapper: Building the VST3 Bundle Folder using the CMAKE_SYSTEM_PROCESSOR variable: (${CMAKE_SYSTEM_PROCESSOR})")
+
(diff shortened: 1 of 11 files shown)


directmusic:
I am new to contributing to open source, so I may have created the branch for this pull request incorrectly. Unfortunately this pull is including commit from the fmt-header-only pull. I am not sure how to correct that.

baconpaul:
OK the way you did this swept in #384 also so let me merge 384 then squash merge this then we can see how it goes. If we want to not use it we can branch around if we need to do a 0.12.2 or something.

Just tagging @defiantnerd so he knows I've done that.

baconpaul:
Oh and @directmusic what you want to do now is reset your next to upstream next.

That's probably something like

```
git remote add upstream https://github.com/free-audio/clap-wrapper
git checkout next
git fetch upstream
git reset upstream/next --hard
```

then your next and repo next will match so subsequent pre will be based properly.

ONLY do that if your next doesn't contain changes.

directmusic:
Thank you so much for the advice! I will be sure to do that

Issue #387, opened 2025-04-22
AU: thread safety issue with request_callback


SamWindell:
My understanding of the CLAP spec is that if a function says [threadsafe] it should be safe to call from any thread - even my own created background thread. This not how the AU version is handling it though. On AU, calling request_callback adds a function to _queueToUI which seems to be a single-producer, single-consumer queue. It needs to be multiple-producer, single-consumer in order to honour the [threadsafe] requirement of request_callback. In reality, I suspect we can just have 2 queues, one for audio-thread to main-thread communication, and another mutex-protected queue for any other thread to main-thread communication and switch based on what thread is calling request_callback.

baconpaul:
but indeed one way to fix it is to have

```
q_type _queueToUI;
q_type _otherThreadToUI;
std::mutex _otherThreadToUILock;
std::thread_id audioThreadID;
```

then do an if this::id == otherwise lock.

SamWindell:
I suspect we might need to go over all thread-safe tagged functions and check we don't make the same assumption.

For example this issues also applies to the thread-check code - which is also tagged [thread-safe] by the CLAP spec. 

In the wrapper code we have a bit of a hack to support thread-check, but it makes the functions completely unreliable to other threads: while the _main/audio_thread_override is incremented then any thread that calls these functions will get true. I think we can just make _main/audio_thread_override a global thread_local variable though then it works.

```cpp
bool Plugin::is_main_thread() const
{
  if (this->_main_thread_override > 0)
  {
    return true;
  }
  return _main_thread_id == std::this_thread::get_id();
}

bool Plugin::is_audio_thread() const
{
  if (this->_audio_thread_override > 0)
  {
    return true;
  }
  return !(_main_thread_id == std::this_thread::get_id());
}
```
https://github.com/free-audio/clap-wrapper/blob/5ba58f16a7816085d46cb8b7bb40b1eb1d9561d2/src/clap_proxy.cpp#L461

defiantnerd:
Yeah, that is a good idea to make that flag thread local. 
And yes, we can't rely on those functions for our own distinction about the locked/not locked queue. Those are completely to transport and satisfy the threading semantics for the actual CLAP.

PR #388, opened 2025-04-23
Fix thread-safety issues related to request_callback

SamWindell:
Following on from #387, here are 2 changes:
- Fix is_main_thread/is_audio_thread by switching to use thread_local. Prior to this they could return true even when not the requested thread.
- Fix race condition in AU wrapper when using request_callback from threads other than the main thread/audio thread at the same time as other thread. We talked about adding a separate queue in the issue. But I found we can just copy what we already do in the VST3 wrapper and it's super simple.

### DIFF
--- a/src/clap_proxy.cpp
+++ b/src/clap_proxy.cpp
@@ -449,9 +449,12 @@ void Plugin::log(clap_log_severity severity, const char* msg)
 #endif
 }
 
+static thread_local uint32_t g_main_thread_override = 0;
+static thread_local uint32_t g_audio_thread_override = 0;
+
 bool Plugin::is_main_thread() const
 {
-  if (this->_main_thread_override > 0)
+  if (g_main_thread_override > 0)
   {
     return true;
   }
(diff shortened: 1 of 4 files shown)


baconpaul:
Oooh why did that override go from a plugin member to a global? That looks very odd to me

SamWindell:
As a thread_local it has a different lifetime to the object now. It could be a static member I suppose but I figured it could just be hidden away in the cpp file.

baconpaul:
right so there's no contention in process because of the locality? can we add a comment to that effect since 'static in plugin' looks like 'bug' to my eyes.

i also wonder if this diff will break the demo i just wrote!

SamWindell:
Right, the purpose of the override is to be able to call into the CLAP's callbacks and if they do a thread-check in that call we return true. We want this to be true for only the current call on the current thread. No one else should see the override so it needs to be thread_local. I'll add a comment for future sake.

Yes the beep demo will be a good test to see if the AU still works correctly! I'll test that.

SamWindell:
Tested with [clap-thread-demo](https://github.com/baconpaul/clap-thread-demo). It's passed the beep test - works the same as before.

PR #391, opened 2025-05-04
Add VST3 Gain Reduction (Presonus Extension)

defiantnerd:
This adds the ability for the VST3 wrapper to readout the extension `CLAP_EXT_GAIN_ADJUSTMENT_METERING` and pass it to VST3 hosts that support the `Presonus::IGainReductionInfo` interface, if the extension is present in the plugin.

### DIFF
--- a/cmake/wrap_vst3.cmake
+++ b/cmake/wrap_vst3.cmake
@@ -31,6 +31,10 @@ function(private_add_vst3_wrapper_sources)
             )
 
     target_include_directories(${V3_TARGET}-clap-wrapper-vst3-lib PRIVATE "${sd}/include")
+
+    # psl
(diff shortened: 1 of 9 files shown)


cmake/wrap_vst3.cmake
```diff
@@ -31,6 +31,10 @@ function(private_add_vst3_wrapper_sources)
             )
 
     target_include_directories(${V3_TARGET}-clap-wrapper-vst3-lib PRIVATE "${sd}/include")
+
+    # psl
+    target_include_directories(${V3_TARGET}-clap-wrapper-vst3-lib PUBLIC libs/psl)
```
baconpaul, line 36: This needs to be “${sd]/libs/psl” to work in subordinate cmake I think (that’s what’s causing the clap first failure)

PR #400, opened 2025-06-01
Issue396, fix crash in AUv2

defiantnerd:
this should communicate an open UI between plugin and wrapped view, also it allows the plugin to notify the view that it is already gone.

I not too happy with this, since I can't exactly replicate the issue itself, also a host is really broken, if such things happen.

### DIFF
--- a/src/detail/auv2/wrappedview.asinclude.mm
+++ b/src/detail/auv2/wrappedview.asinclude.mm
@@ -19,6 +19,7 @@
 @interface CLAP_WRAPPER_COCOA_CLASS_NSVIEW : NSView
 {
   free_audio::auv2_wrapper::ui_connection ui;
+  uint32_t canary;
   CFRunLoopTimerRef idleTimer;
   float lastScale;
   NSSize underlyingUISize;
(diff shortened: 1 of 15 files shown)


src/wrapasauv2.cpp
```diff
@@ -147,6 +148,11 @@ WrapAsAUV2::~WrapAsAUV2()
     _os_attached.off();
     _plugin->terminate();
     _plugin.reset();
+    if (_uiIsOpened && _uiconn._canary)
```
baconpaul, line 151: I think it might help to comment this case. This is the case where plugin is destroyed but ui object still is not? Why do we not need to do anything else here like gui destroy?

src/detail/auv2/wrappedview.asinclude.mm
```diff
@@ -19,6 +19,7 @@
 @interface CLAP_WRAPPER_COCOA_CLASS_NSVIEW : NSView
 {
   free_audio::auv2_wrapper::ui_connection ui;
+  uint32_t canary;
```
baconpaul, line 22: Are we sure canary or the pointers to canary don’t need to be atomic?

src/detail/auv2/wrappedview.asinclude.mm
```diff
@@ -150,9 +154,12 @@ - (void)dealloc
   {
     CFRunLoopTimerInvalidate(idleTimer);
   }
-  auto gui = ui._plugin->_ext._gui;
-  gui->destroy(ui._plugin->_plugin);
-
+  if ( canary )
+  {
+    auto gui = ui._plugin->_ext._gui;
```
baconpaul, line 159: Do we need to lock in these if canary blocks generally? If the theory is that arc/gc leaves the window linger are we sure it reaps on the same thread as the plugin destroy and if not could we have a race here absent destroy lock?

src/wrapasauv2.cpp
defiantnerd: you're right - we need to destroy the UI here ourselves

defiantnerd:
Plugin gui destruction has now been moved to `viewDidMoveToWindow()` when the parenting window is nil.
I checked against AULab, Logic, Live, FL Studio, Bitwig and Reaper, all conform this destruction sequence.

PR #402, opened 2025-06-08
add  CLAP_TRANSPORT_HAS_SECONDS_TIMELINE and  CLAP_TRANSPORT_HAS_BEAT…

defiantnerd:
When filling the members `song_pos_seconds`and `song_pos_beats` the flags `CLAP_TRANSPORT_HAS_SECONDS_TIMELINE` and  `CLAP_TRANSPORT_HAS_BEATS_TIMELINE` need to be added to the process flags, too.

### DIFF
--- a/src/detail/standalone/standalone_host.cpp
+++ b/src/detail/standalone/standalone_host.cpp
@@ -110,8 +110,9 @@ void StandaloneHost::clapProcess(void *pOutput, const void *pInput, uint32_t fra
   }
 
   auto f = (float *)pOutput;
+
   clap_process process;
-  process.transport = nullptr;
+  process.transport = nullptr;            // this is a freefloating host
   process.in_events = &inputEvents;
   process.out_events = &outputEvents;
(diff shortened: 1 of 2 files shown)


src/detail/clap/fsutil.cpp
```diff
@@ -195,6 +195,11 @@ bool Library::load(const fs::path &path)
       _handle = nullptr;
     }
   }
+  else
+  {
+    auto err = ::GetLastError();
+    OutputDebugString("ups");
```
baconpaul, line 201: Do we want to show the error here rather than ups

baconpaul (commented):
Looks good except the error thing

src/detail/clap/fsutil.cpp
defiantnerd: I have this bad habit of using `git stage -u`... that was supposed to be a different fix - now it is in here, just added some more handling.

PR #403, opened 2025-06-27
Handle NULL in `clap_audio_port_info.port_type`

geraintluff:
The `std::strcmp` segfaults if given a NULL string.


defiantnerd:
Not sure if we should put this to next instead of main.

baconpaul:
we should always go to next before main

geraintluff:
Bump

defiantnerd:
Thanks!

PR #405, opened 2025-07-17
fix #392 for VST3

### LINKED ISSUE #392: Null pointer dereference if create_plugin() returns null
CLAP plugins are allowed to return null from create_plugin if there's an error: https://github.com/free-audio/clap/blob/69a69252fdd6ac1d06e246d9a04c0a89d9607a17/include/clap/factory/plugin-factory.h#L30-L37

The wrapper should make an effort to handle this.

SamWindell:
Fix null dereference when createInstance has an error and returns nullptr.

### DIFF
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -129,7 +129,7 @@ tresult PLUGIN_API ClapAsVst3::initialize(FUnknown* context)
     {
       _plugin = Clap::Plugin::createInstance(*_library, _libraryIndex, this);
     }
-    result = (_plugin->initialize()) ? kResultOk : kResultFalse;
+    result = (_plugin && _plugin->initialize()) ? kResultOk : kResultFalse;
   }
 
   return result;


defiantnerd:
this was fixed with https://github.com/free-audio/clap-wrapper/pull/401 in `next`

SamWindell:
With #401 it's only for the clap proxy side and AU side, not VST3. 

This PR fixes the same issue if it happens in a VST3 (where a problem in the CLAP results in null being returned). I had this come up in a real world situation.

I understand that if a CLAP is returning null it's a problem. But it's just good damage mitigation to still try and handle these errors rather than crash by null dereference.

defiantnerd:
it is solved generally at this point: 
https://github.com/free-audio/clap-wrapper/blob/48e3859ab2494bcfcabb874b76c5693a3c6ce5c8/src/clap_proxy.cpp#L144

If the clap does not return an plugin instance pointer, this is handled and the plug resets itself.

SamWindell:
I swear I got a stacktrace of a null dereference crash from a user on the line I changed but I can't find it at the moment!

And looking closer, I think you are right, instance can't be null.

But... actually, there's something else here. My change is still necessary. 

Look, createInstance can return null when `library.plugins.size() <= index`:
https://github.com/free-audio/clap-wrapper/blob/48e3859ab2494bcfcabb874b76c5693a3c6ce5c8/src/clap_proxy.cpp#L164-L182

And then the caller of createInstance, ClapAsVst3::initialize, dereferences it:
https://github.com/free-audio/clap-wrapper/blob/48e3859ab2494bcfcabb874b76c5693a3c6ce5c8/src/wrapasvst3.cpp#L122-L136

PR #411, opened 2025-07-30
Fix thread-safety when calling request_resize()

defiantnerd:
This patch checks if calls to request_resize() happen on the main UI thread, and if not, sets a flag and stores the requested size so ClapAsVst3::onIdle() can handle the request.

### DIFF
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -1111,7 +1111,7 @@ bool ClapAsVst3::gui_can_resize()
 bool ClapAsVst3::gui_request_resize(uint32_t width, uint32_t height)
 {
   // UIs with 65kx65k resolution are not supported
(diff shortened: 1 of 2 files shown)


defiantnerd:
@SamWindell @directmusic 
I refactored the PR 407 to be cleaner. What do you think?

src/wrapasvst3.cpp
```diff
@@ -1106,6 +1110,16 @@ bool ClapAsVst3::gui_can_resize()
 
 bool ClapAsVst3::gui_request_resize(uint32_t width, uint32_t height)
 {
+  // UIs with 65kx65k resolution are not supported
+  if ((width > 0xffff) && (height > 0xffff)) return false;
```
SamWindell, line 1114: Should be `||`

Issue #395, opened 2025-05-09
Avoid concurrent main-thread callbacks with AU Logic Pro


SamWindell:
Logic Pro using an AU wrapped CLAP is calling `GetParameterInfo` from the audio thread (when I query the name of the thread it returns `com.apple.audio`...). 

In GetParameterInfo, the wrapper calls the CLAP's params.get_value function - main thread only.

It's only doing this to set a flag though, we could avoid this by querying this once earlier on rather than in the GetParameterInfo callback

https://github.com/free-audio/clap-wrapper/blob/5ba58f16a7816085d46cb8b7bb40b1eb1d9561d2/src/wrapasauv2.cpp#L457-L467

baconpaul:
we can clearly at setup time use value_to_text with the default value indeed at setup time. (value_to_text is main thread only also iirc)

SamWindell:
Just a little more info on this:

In my temporary fork I have commented out the code that calls get_value and value_to_text, and just always add the kAudioUnitParameterFlag_HasName flag instead. I've been using this for a few weeks now and the CLAP wrapper behaves correctly in Logic and Garageband: there are no long parallel 'main thread' calls.

baconpaul:
Great! So it really does sound like thats the problem then?

defiantnerd:
So actually, this is no problem at all since it just gathers rendered text strings for automation and generic view.

A plugin should be consistent enough to handle this.

It is also a possibility to declare this "must be thread safe", isn't it?

SamWindell:
I still think there's a problem here, and an easy fix.

Here are the facts:
1. Logic Pro and Garage Band call AU `GetParameterInfo` from the audio thread.
2. Within `GetParameterInfo`, clap-wrapper is calling the CLAP's `get_value` and `value_to_text`: these 2 functions are main-thread according to CLAP spec.
3. The result is that the CLAP can receive parallel callbacks from functions that are meant to be main-thread according the CLAP spec.

So my solution: don't call `get_value` or `value_to_text` from inside `GetParameterInfo`. Instead it can be cached earlier, or just assume always kAudioUnitParameterFlag_HasName.


You could claim Logic/Garageband are broken for calling `GetParameterInfo` from the audio thread. But who's so say what counts as 'broken' for AU? As far as I know, there is no clear spec on what threads are used for what.


[...]

PR #412, opened 2025-07-31
AUv2: make GetParameterInfo() thread safe

defiantnerd:
should fix an AUv2 issue when GetParameterInfo() is being called
See #395


baconpaul:
This looks correct to me. Are you happy to merge it?

PR #415, opened 2025-08-28
Feature/auv2 musiceffect support

davemollen:
I want to create an AUv2 plugin of type "aumf" (music effect) when the MACOSX_EMBEDDED_CLAP_LOCATION is set. 

In the first commit I made INSTRUMENT_TYPE take precedence over setting the plugin type based on the present CLAP features. This way you could also create all the other possible AUv2 plugin types.

In the second commit I changed the automatic AUv2 plugin type assignment a bit. When the CLAP plugin has both the AudioEffect and the Instrument feature the "aumf" plugin type is assigned. Also, this checks the whole set of features, instead of only the first.

I wouldn't be surprised if this not mergeable yet. Maybe I'm missing some caveats. But I'd be happy to contribute. I could also turn this into an issue if this is way off.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ ignore/
 build/
 out/
 .DS_Store
+
+#vscode
+.vscode/*
--- a/cmake/wrap_auv2.cmake
+++ b/cmake/wrap_auv2.cmake
@@ -88,7 +88,6 @@ function(target_add_auv2_wrapper)
         get_property(ton TARGET ${clpt} PROPERTY LIBRARY_OUTPUT_NAME)
         set(AUV2_OUTPUT_NAME "${ton}")
         set(AUV2_SUBTYPE_CODE "Fooo")
-        set(AUV2_INSTRUMENT_TYPE "aumu")
 
         add_dependencies(${AUV2_TARGET} ${clpt})
         add_dependencies(${bhtg} ${clpt})
@@ -107,7 +106,6 @@ function(target_add_auv2_wrapper)
     elseif (DEFINED AUV2_MACOSX_EMBEDDED_CLAP_LOCATION)
         message(STATUS "clap-wrapper: building auv2 based on clap ${AUV2_MACOSX_EMBEDDED_CLAP_LOCATION}")
         set(AUV2_SUBTYPE_CODE "----")
-        set(AUV2_INSTRUMENT_TYPE "aumu")
 
         add_custom_command(
                 TARGET ${bhtg}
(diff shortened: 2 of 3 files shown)


defiantnerd:
Hey, thanks for your input. I'll check it out tomorrow.

baconpaul:
Can’t you do this instead with the au extension which lets you set your type explicitly? 

https://github.com/free-audio/clap-wrapper/blob/62bf4f193d8dac354a946c1e0cffda9174317042/include/clapwrapper/auv2.h#L16

If your clap implements that extension it should use it in the compile time generation of the plist

davemollen:
I tried setting the type explicitly by passing INSTRUMENT_TYPE to the target_add_auv2_wrapper function. But that doesn't work when I also use the MACOSX_EMBEDDED_CLAP_LOCATION. Otherwise it works fine.
Should I pass the audio unit type in some other way to set it explicitly?

baconpaul:
Hmm that embedded option must use another path to make the plist indeed. I’ll try and look

baconpaul:
Hi

OK sorry for the delay here. Which is very long I know!

I spent some time looking at this today. I had no problem at all using the auv2 extension with an embedded clap to set the instrument type to aumf

I merged my test example here in the 'tests' directory in next

But I think your problem is using cmake + embedded location yeah? If you use the in-clap extension it all works.

Let me look at that next now I have an example working, but if you happen to be online and can make the example look like your code or at least eyeball the difference that would help.

baconpaul:
Ahh now I see the problem. If you have the cmake specified *and* have the embedded target, it uses the embedded target, finds no extension, and defaults it.

Gotcha thats a cmake switch. Stay tuned.

baconpaul:
OK so now I see the problem - if you have an embedded clap that always 'wins' for information resolution.

That's what we want in many cases but not all. 

So in https://github.com/free-audio/clap-wrapper/pull/447 I added a new wrap auv2 option PREFER_CMAKE_AUV2_CONFIGURATINO which means if you have the arguments and the embedded clap then the arguments beat the embedded clap in all cases.

That's the right solution here I think. It let me set a type to qewe and it all worked. I'll close this and merge that to next for our .14 release.

Thanks for your patience.

baconpaul:
Ahh actually even better I can make the prefer automatic with one more tricky commit. Stay tuned

davemollen:
That's great, I look forward to trying this out!
And I'll stay tuned for the automatic assignment too. That would be even more convenient

baconpaul:
Yeah the automatic is at head of next

Basically if you have an embedded clap and cmake args, the cmake args win unless your embedded clap implements the auv2 extension. And if you have neither the args nor the extension it will revert to old behavior.

davemollen:
I finally got round to testing this. It works great, thanks so much!

PR #417, opened 2025-09-25
Implement host request_restart on standalone

pkhead:
This PR fully implements the clap_host_t request_restart method for Windows and macOS standalone, although the macOS implementation is currently untested.

### DIFF
--- a/src/detail/standalone/standalone_host.cpp
+++ b/src/detail/standalone/standalone_host.cpp
@@ -103,13 +103,15 @@ void StandaloneHost::setupMIDIBusses(const clap_plugin_t *plugin,
 
 void StandaloneHost::clapProcess(void *pOutput, const void *pInput, uint32_t frameCount)
 {
+  ClapWrapper::detail::shared::SpinLockGuard processLockGuard(processLock);
+  auto f = (float *)pOutput;
+  
   if (!running)
   {
+    memset(f, 0, frameCount * 2 * sizeof(float));
     finishedRunning = true;
     return;
   }
-
-  auto f = (float *)pOutput;
 
   clap_process process;
   process.transport = nullptr;  // this is a freefloating host
--- a/src/detail/standalone/standalone_host.cpp
+++ b/src/detail/standalone/standalone_host.cpp
@@ -105,7 +105,7 @@ void StandaloneHost::clapProcess(void *pOutput, const void *pInput, uint32_t fra
 {
   ClapWrapper::detail::shared::SpinLockGuard processLockGuard(processLock);
   auto f = (float *)pOutput;
-  
+
   if (!running)
   {
     memset(f, 0, frameCount * currentOutputChannels * sizeof(float));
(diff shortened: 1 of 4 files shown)


defiantnerd:
Thanks. Unfortunately this is not enough, audio/midi must be stopped during the restart. Your PR will probably lead to crashes and also they should trigger clap-helpers sanity checks.

baconpaul:
why do you say that timo?

my concern about this was more that the activate doesn't wait for an inflight process to finish after setting audio running

so basically you do

sah->running = false
activate
sah->running = true

but if there's a process running with a large block size, that activate may happen before the process finishes and loops around to check running again

defiantnerd:
We do mean the same thing. The host must make sure that it is not in a process call and also does not enter it whilst re-activating.

Therefore the processing must be stopped or bypassed.

baconpaul:
right so this will need to do a little spin and process will have to set a skipped atomic.

pkhead:
oh. right. silly me. i was certain that a bug or oversight existed in my code, as i only skimmed the codebase for like 20 minutes before making this pr. i made this quick change to see if it worked, and i honestly i was uncertain if i should make an issue or a pr.

i did notice the existence of that fairly obvious unaddressed concurrency problem some time after submitting this. probably what's causing that weird audio glitch on my end.

er, i can try to implement the spinlock. i saw spinlock code somewhere in this project, i suppose i could find it and reappropriate it for usage here.

also, while audio is addressed i'm not sure if midi processing should be a concern. is it or is it not? or does implementing this fix fix whatever problems might occur with midi?

baconpaul:
luckily in clap all events run through process so if you get that locked off properly then indeed midi note expression etc... will all be taken care of.

pkhead:
i think this is okay

src/detail/standalone/standalone_host.cpp
```diff
@@ -103,14 +103,16 @@ void StandaloneHost::setupMIDIBusses(const clap_plugin_t *plugin,
 
 void StandaloneHost::clapProcess(void *pOutput, const void *pInput, uint32_t frameCount)
 {
+  ClapWrapper::detail::shared::SpinLockGuard processLockGuard(processLock);
+  auto f = (float *)pOutput;
+  
   if (!running)
   {
+    memset(f, 0, frameCount * 2 * sizeof(float));
```
defiantnerd: instead of the magic number "2" here should be the actual number of channels even if the current standalone implementation only assumes stereo.

defiantnerd:
@baconpaul I would be okay with that for now, WDYT?

defiantnerd:
@pkhead please check on the failed workflow issues. If you don't have time, let me know.

baconpaul:
I pushed up a change to the clang format and misnamed macOS variable in AppDelegate.mm. Lets see how CI does now.

baconpaul:
I also changed the base branch to next

defiantnerd:
thanks a lot - will merge to next

PR #419, opened 2025-11-01
VST3: Make host->request_restart thread-safe

pkhead:
This uses the same method as request_callback to make request_restart thread-safe. The restart will chronologically occur after the UI callback request is handled.


baconpaul:
@pkhead I'm sorry I missed this.

@defiantnerd if we are getting ready for a 0.14 should we sweep this one? Thoughts?

defiantnerd (approved):
Yes, totally reasonable change. thanks!

PR #421, opened 2025-11-21
handle common behaviour between emscripten/wasi toolchains

Trinitou:
saw some inconsistencies regarding wasm:
- correct fallback for `PLUGIN_FORMATS` not set in case of wasm
- standalone isn't implemented for wasi

this PR tries to fix them. not really tested yet though

side note: I haven't touched `CLAP_WRAPPER_PLATFORM` because I'm unsure what to do with it. In case of wasi it is not set which looks wrong.


geraintluff:
Looks great, +1 from me

PR #422, opened 2025-12-08
Change WCLAP max-memory 4GB -> 2GB

geraintluff:
2GB is the default Emscripten uses, for reasons explained [here](https://emscripten.org/docs/tools_reference/settings_reference.html#maximum-memory):

> Note that the default value here is 2GB, which means that by default if you enable memory growth then we can grow up to 2GB but no higher. 2GB is a natural limit for several reasons:
> * If the maximum heap size is over 2GB, then pointers must be unsigned in JavaScript, which increases code size. We don’t want memory growth builds to be larger unless someone explicitly opts in to >2GB+ heaps.
> * Historically no VM has supported more >2GB+, and only recently (Mar 2020) has support started to appear. As support is limited, it’s safer for people to opt into >2GB+ heaps rather than get a build that may not work on all VMs.

This change matches that when compiling with WASI SDK, providing more consistent WCLAP results.


defiantnerd:
Sounds reasonable for me.

PR #424, opened 2026-01-10
Add VST3 Validator + auval steps to the pull request validation workflow

Quant1um:
This PR adds a VST3 validator test step and an auval step on macOS when building the ClapFirst example plugin. Currently it builds and runs the official VST3 validator provided by Steinberg.

Also this PR fixes a small issue with VST3 bundle folder generation on Windows where clap-wrapper used a wrong architecture name, this is needed for the validator to pass.

Tested that it works on my fork [here](https://github.com/Quant1um/clap-wrapper/actions/runs/20881660854)

### DIFF
--- a/cmake/wrap_vst3.cmake
+++ b/cmake/wrap_vst3.cmake
@@ -216,12 +216,17 @@ function(target_add_vst3_wrapper)
 
             # Check against the list of supported targets found here:
             # https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Locations+Format/Plugin+Format.html#for-the-windows-platform
-            if(NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "x86"
-                    AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64"
-                    AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64ec"
-                    AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64"
-                    AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64x")
+            if (CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
+                set(v3arch "x86_64")  
+            elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "IA64")
+                set(v3arch "x86_64")
+            elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86")
+                set(v3arch "x86")
+            elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "ARM64")
+                set(v3arch "arm64")
+            else()
                 message(WARNING "clap-wrapper: The architecture (${CMAKE_SYSTEM_PROCESSOR}) is not officially suported by VST3. This may cause issues when loading the resulting plug-in")
+                set(v3arch "${CMAKE_SYSTEM_PROCESSOR}")
             endif()
 
             if ("${V3_ASSET_OUTPUT_DIRECTORY}" STREQUAL "")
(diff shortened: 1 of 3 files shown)


cmake/wrap_vst3.cmake
```diff
@@ -216,12 +216,15 @@ function(target_add_vst3_wrapper)
...
-                    AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64ec"
-                    AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64"
-                    AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64x")
+            if (CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
+                set(v3arch "x86_64")
+            elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "X86")
+                set(v3arch "x86")
+            elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "ARM64")
+                set(v3arch "arm64")
+            else()
```
baconpaul, line 225: Can you help me understand this change? Did we just have it wrong before?

baconpaul: Oh i see in your comment. Yes, we just had it wrong.

baconpaul:
OK this is super duper useful thank you. I'm happy to merge it.

@defiantnerd - good for you?

defiantnerd:
I actually would like that it would use the already downloaded SDK. So we can keep the same tag of it.

Also, it should not enforce the "plugin is a folder" ob Windows. I am sure this is not an issue for the validator.

But apart of that: yes, super useful.
Thank you!

baconpaul:
The plugin is a folder is just for the example and apparently is required for the validator according to the comment here - I haven’t tried myself.  But the fix to the architecture names is good 

I’ll take a look at how to reliably find the name of the downloaded version in the download build - it should have a stable name but I don’t know if we will know it easily outside of our cmake; and since our cmake doesn’t use the steinberg cmake it can’t build the validator

Quant1um:
> I actually would like that it would use the already downloaded SDK. So we can keep the same tag of it.
> 
> Also, it should not enforce the "plugin is a folder" ob Windows. I am sure this is not an issue for the validator.
> 
> But apart of that: yes, super useful. Thank you!

It seems that the official validator does require the plugin to be a folder on Windows, otherwise it loads fine but the validator shows one of the tests as failed

defiantnerd:
That would be a pity ... I'll check it out.

defiantnerd:
ok, as we can see here, loading as DLL is still supported:
https://github.com/steinbergmedia/vst3_public_sdk/blob/a3911a4615dabbfdfd9d181ee26b05c70c289a95/source/vst/hosting/module_win32.cpp#L205

And it works, I've just checked it locally with the 3.8.0 version of the SDK. 
Can we work on this? It is probably just a path issue.

Quant1um:
The error I get when running the validator workflow is:
```
> Run "./vst3sdk/cmake-build/bin/Debug/validator" "./build/clap-first-distortion_assets/VST3/Debug/ClapFirst Bad Distortion.vst3"
* Loading module...

	./build/clap-first-distortion_assets/VST3/Debug/ClapFirst Bad Distortion.vst3

* Check valid bundle structure...
Invalid bundle structure: Not a bundle: './build/clap-first-distortion_assets/VST3/Debug/ClapFirst Bad Distortion.vst3'.

...

 -------------------------------------------------------------
Result: 47 tests passed, 1 tests failed
-------------------------------------------------------------
```

It seems to load just fine, but instead it fails [here](https://github.com/steinbergmedia/vst3_public_sdk/blob/a3911a4615dabbfdfd9d181ee26b05c70c289a95/samples/vst-hosting/validator/source/validator.cpp#L474).

defiantnerd:
ok, currently, clap-wrapper sticks with 3.7.6 and the validator in that does not check for modules. Perhaps it is a bug in the validator. I'll checkout the 3.8.0 version.

defiantnerd:
With master/3.8.0 it works as well. Which version of the VST3 sdk are you using?

Quant1um:
I'm using the master (3.8.0) version. The error above is the result of running the Github Action on my own fork of clap-wrapper.
Using the 3.7.11 SDK validator seems to work though. It's probably a good idea to open a new issue in the vst3sdk repo.

defiantnerd:
Does it work locally on your machine? Ot does on mine. Maybe it is an issue with the github action infrastructure.

baconpaul:
OK here's what I'm gonna do this morning. I'll pull this branch and then see if I can write a cmake rule to build the validator using the same version as the wrapper so I can simplify the actions a bit. 

@Quant1um i'll push this to your branch rather than start a new PR unless you object to that.

Quant1um:
@defiantnerd Exact same thing happens when testing locally.

@baconpaul Sounds good!

baconpaul:
Ahh actually you are on your next not on a branch. Makes it a bit clumsy. But stay tuned. I may end up opening a different PR with a named branch after all.

baconpaul:
OK using the built in vst3 sdk i get quite a different result and build than you do (including build failure on gcc12). you can se the other pr here.

I also tried to update to vst3 3.8.0 for ourselves and got build failures in the wrapper.

So let me try and get to 3.8.0 separately on next first.

baconpaul:
Alright in #425 I have it working on mac and win at least using the chosen SDK rather than an extra download and build
I'll see if I can make #426 work (move to 3.8.0) first and then rebase that and see if I can get linux back also.
stay tuned.

baconpaul:
@defiantnerd over in #425 I have this all put together

but I also did a folder-on-folder-off case and the folder-off case on windows clearly fails to validate

https://github.com/free-audio/clap-wrapper/actions/runs/20901158923/job/60047231771?pr=425 there's the run

and I get 

```
Run "build/validator-build/bin/validator" "./build/clap-first-distortion_assets/VST3/ClapFirst Bad Distortion.vst3"
* Loading module...

	./build/clap-first-distortion_assets/VST3/ClapFirst Bad Distortion.vst3

Invalid Module!
LoadLibraryW failed for path ./build/clap-first-distortion_assets/VST3/ClapFirst Bad Distortion.vst3: The specified module could not be found.
```

I modified my PR to only run the validator on the bundle version and it passed (https://github.com/free-audio/clap-wrapper/actions/runs/20901288511/job/60047560096?pr=425)

So I think @Quant1um is correct. The 3.8 validator at least demands a module on windows.

baconpaul:
Also @Quant1um i'd rather close this in favor of #425 which includes your code but moves the validator build into cmake rather than standalone separate using the same version etc....

Quant1um:
> Also @Quant1um i'd rather close this in favor of #425 which includes your code but moves the validator build into cmake rather than standalone separate using the same version etc....

@baconpaul I'm fine with that!

baconpaul:
great #425 merged (with you as a co-author)

Thanks! Great suggestion.

defiantnerd:
And thanks @Quant1um!

PR #428, opened 2026-01-16
Fix WCLAP build

geraintluff:
The `ANY_WASM_TOOLCHAIN` variable set in #421 is only set in the local scope, not `PARENT_SCOPE`, meaning that it still attempts to build VST3/etc. instead of WCLAP.

This change fixes that, as well as setting `CLAP_WRAPPER_PLATFORM` appropriately.


baconpaul (approved):
Looks good to me. OK with you @defiantnerd ?

PR #429, opened 2026-01-16
Ableton VST3 crash: `NULL` strings in plugin descriptor

geraintluff:
In `clap_plugin_descriptor`, `vendor`/`url`/`version` are not mandatory, meaning they _can_ be `NULL`.

The VST3 wrapper doesn't check for this, leading to a crash at least some DAWs (confirmed in Ableton).

### DIFF
--- a/src/wrapasvst3_entry.cpp
+++ b/src/wrapasvst3_entry.cpp
@@ -334,12 +334,14 @@ IPluginFactory* GetPluginFactoryEntryPoint()
             n.append(" (CLAP->VST3)");
 #endif
             auto plugname = n.c_str();  //  clapdescr->name;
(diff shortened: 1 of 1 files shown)


baconpaul:
I also think this one is good, but I always get a bit shaky when i see const char* assigned to literals and while this looks correct, let me ask @defiantnerd this is fine right? Second set of eyes just to make sure I'm not misreading.

defiantnerd (approved):
I can't imagine why anyone would pass a nullptr there, but yes, it is allowed in CLAP and not in VST3.

defiantnerd:
@geraintluff thanks for that, I'd say we can merge that.

PR #430, opened 2026-01-16
Implement several multi-layout extensions for VST3

Quant1um:
This PR implements:
- Multi-layout support for VST3 backend via `configurable-audio-ports` and `audio-ports-config` (whichever is present)
- Bus activation flags via `audio-ports-activation`
- Better support for null-typed ports with arbitrary number of channels (`speakerArrFromPortType`)
- `audio-ports-config` implemented for `clap-first-example` so it could be validated

A similar feature probably needs to be implemented on the AUv2 side, but I don't have enough experience working with AUv2.

### DIFF
--- a/src/detail/shared/util.h
+++ b/src/detail/shared/util.h
@@ -1,8 +1,25 @@
 #pragma once
 
 #include <cstdint>
+#include <string>
+
+#if __cplusplus >= 202002L
+#include <bit>
+#endif
 
 uint32_t fnv1a_keogh(const char *input);
 
 // Function to shorten a string to a 4-character string
 std::string ShortenString(const std::string &input);
+
+inline uint32_t popcount64(uint64_t x)
+{
+#if __cplusplus >= 202002L
+  return std::popcount(x);
+#else
+  x = x - ((x >> 1) & 0x5555555555555555ULL);
+  x = (x & 0x3333333333333333ULL) + ((x >> 2) & 0x3333333333333333ULL);
+  x = (x + (x >> 4)) & 0x0F0F0F0F0F0F0F0FULL;
+  return (x * 0x0101010101010101ULL) >> 56;
+#endif
+}
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -22,6 +22,7 @@
 #include "detail/vst3/process.h"
 #include "detail/vst3/parameter.h"
 #include "detail/clap/fsutil.h"
+#include "detail/shared/util.h"
 #include <locale>
 #include <sstream>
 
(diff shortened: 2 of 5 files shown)


baconpaul:
OK this is a big change. We may want to drop 0.14 first (planned this week) and then put this on next after that.

I also note that your diff changes `const char *foo` to `const char* foo` throughout making a lot of diff noise. Any reason fro that? I'm honestly a bit surprised it passed clang format.

src/wrapasvst3.cpp
```diff
@@ -65,6 +65,18 @@ struct ClapHostExtensions
 };
 #endif
 
+// is there a better way?
```
defiantnerd: You can use this
```c++
uint32_t popcount64(uint64_t x)
{
    x = x - ((x >> 1) & 0x5555555555555555ULL);
    x = (x & 0x3333333333333333ULL) + ((x >> 2) & 0x3333333333333333ULL);
    x = (x + (x >> 4)) & 0x0F0F0F0F0F0F0F0FULL;
    return (x * 0x0101010101010101ULL) >> 56;
}
```
or with c++20:
```c++
#include <bit>

uint32_t popcount64(uint64_t x) {
    return std::popcount(x);
}
```
Perhaps we can put this to utils.

Quant1um:
@baconpaul The formatting change was not intentional, just fixed it! I think it would be a good idea to add `PointerAlignment: Right` (or `Left`) to `.clang-format`, currently pointer alignment seems to be all over the place.

src/wrapasvst3.cpp
Quant1um: > Perhaps we can put this to utils.

Is there a file I can move this function to?

defiantnerd: how about \src\detail\shared\util.h
We could inline it and distinguish between C++20 or less. What do you think?

defiantnerd:
Oh, and `audio-ports-config` is not suitable for this - the only reason for this extension is a list to be presented to the user. But the settings do not contain enough information so setup the specific busses when it comes to surround/ambisonic.

src/wrapasvst3.cpp
Quant1um: Done!

Quant1um:
Okay, I removed `audio-ports-config` extension support; there are still some unresolved questions:
- VST3's `setBusArrangements` wants the plugin to report a _suggested_ configuration if the plugin doesn't accept the given config. Current implementation defaults to suggesting the current plugin layout. What would the desired behavior be?
- Should this PR also implement 'surround' extension and report layouts with more than 2 channels as CLAP_PORT_SURROUND, or should this be left to a future PR? (currently it reports them as null-typed ports)

defiantnerd:
I would think ghat suggested config is the current default config - which can be determined from audio-ports. Maybe I am thinking too simple.

src/detail/shared/util.h
```diff
@@ -0,0 +1,17 @@
+
+#pragma once
+
+#include <cstdint>
+#include <bit>
```
defiantnerd, line 7: <bit> is only available in c++20, I guess this must be within the #if clause

Quant1um: Oops. Fixed!

defiantnerd:
Thanks! Very good contribution!

Quant1um:
Just did a rebase, it should be mergeable now

baconpaul:
Thanks. We decided to hold this until *after* .14 for a couple of reasons

1. We just want to test it with a bunch of multi-out plugins
2. It will interact with AAX anyway which is post .14

So while we don' have explicit changes to propose now we're going to wait until we stamp stable to merge this.

PR #431, opened 2026-01-21
Fix VST3 memory leaks

Quant1um:
I found that when validating the example plugin with ASAN on the wrapper leaks about 2mb of memory. This PR fixes it by:
- Using `std::unique_ptr` where applicable, and
- Removing `result->addRef();` in `Vst3Parameter::create`. Not sure if it is correct, or what the original purpose of this was, but it seems to work in validator under ASAN, Bitwig and Reaper. Review needed.

Side note: it would probably be a good idea to turn on address sanitizer while validating the plugin in CI

### DIFF
--- a/src/detail/vst3/parameter.cpp
+++ b/src/detail/vst3/parameter.cpp
@@ -132,7 +132,7 @@ Vst3Parameter* Vst3Parameter::create(
     v.stepCount = 0;
 
   auto result = new Vst3Parameter(v, info);
-  result->addRef();  // ParameterContainer doesn't add the ref -> but we don't have copies
+  //result->addRef();  // ParameterContainer doesn't add the ref -> but we don't have copies
   return result;
 }
 
@@ -184,6 +184,6 @@ Vst3Parameter* Vst3Parameter::create(uint8_t bus, uint8_t channel, uint8_t cc, V
   }
 
   auto result = new Vst3Parameter(v, bus, channel, cc);
-  result->addRef();  // ParameterContainer doesn't add the ref -> but we don't have copies
+  //result->addRef();  // ParameterContainer doesn't add the ref -> but we don't have copies
   return result;
 }
--- a/src/detail/vst3/process.cpp
+++ b/src/detail/vst3/process.cpp
@@ -34,23 +34,19 @@ void ProcessAdapter::setupProcessing(const clap_plugin_t* plugin, const clap_plu
 
   if (numSamples > 0)
   {
-    delete[] _silent_input;
-    _silent_input = new float[numSamples];
-
-    delete[] _silent_output;
-    _silent_output = new float[numSamples];
+    // wait do we even use these?
+    _silent_input = std::make_unique<float[]>(numSamples);
+    _silent_output = std::make_unique<float[]>(numSamples);
   }
 
   auto numInputs = (uint32_t)_audioinputs->size();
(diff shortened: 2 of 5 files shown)


baconpaul:
I'm pretty sure removing the addRef is wrong, and instead we should add a removeRef when we are done with our parameters explicitly. ASAN in the validator wouldn't necessarily catch this since the validator holds a reference to the param while using it and doesn't try to get it again, as far as I know.

Quant1um:
> I'm pretty sure removing the addRef is wrong, and instead we should add a removeRef when we are done with our parameters explicitly. ASAN in the validator wouldn't necessarily catch this since the validator holds a reference to the param while using it and doesn't try to get it again, as far as I know.

Hmm, it is my understanding that `ParameterContainer::addParameter` takes ownership of the passed `Parameter` (it uses the IPtr constructor that doesnt add to refcount, so it stays 1 after `addParameter` and it gets released in `ParameterContainer::removeAllParams` and the destructor). The official [again](https://github.com/steinbergmedia/vst3_public_sdk/blob/a3911a4615dabbfdfd9d181ee26b05c70c289a95/samples/vst/again/source/againsimple.cpp#L136) example seems to be consistent with this.

Also I am not sure that hosts even interact with the parameter reference (instead they use the getParameterInfo/getParamNormalized/etc API which uses parameters IDs to refer to them)

Correct me if I'm wrong

baconpaul:
Right so you are saying the code *and * the comment are wrong! Got it.

@defiantnerd - you have git attribution for that addRef. Thoughts?

The unique ptr changes all look great to me! It's just that addRef I want to have someone else eye up.

Thanks!

baconpaul:
Sorry for the delay. Timo and I hopped on a call and looked at this in detail and you are exactly correct. Merging.

baconpaul:
Hi

I reverted this commit today. With it, you cant open and close short circuit twice in reaper

I think the problem is the change of wrapped view from a * to a unique ptr means unique ptr and the steinberg funknown are fighting with each other 

I'll rebuild it section by section before we do 0.14, which we've now decided to delay.

baconpaul:
```
0   libsystem_kernel.dylib                   0x19b3c95f0 __pthread_kill + 8
1   libsystem_pthread.dylib                  0x19b401c20 pthread_kill + 288
2   libsystem_c.dylib                        0x19b30ea30 abort + 180
3   libsystem_malloc.dylib                   0x19b21edc4 malloc_vreport + 896
4   libsystem_malloc.dylib                   0x19b222430 malloc_report + 64
5   libsystem_malloc.dylib                   0x19b23c494 find_zone_and_free + 528
6   Shortcircuit XT                          0x31074b7f4 WrappedView::~WrappedView() + 36 (plugview.cpp:20)
7   Shortcircuit XT                          0x311b45f3c Steinberg::FObject::release() + 156 (fobject.cpp:127)
8   Shortcircuit XT                          0x31074f570 WrappedView::release() + 28 (plugview.h:92)
9   REAPER                                   0x100ee01d4 VST_HostedPlugin::VST3_Dispatcher(AEffect*, int, int, long long, void*, float) + 3492
10  REAPER                                   0x100d1a6dc VST_HostedPlugin::QuitConfig() + 300
11  REAPER                                   0x100d27190 FxDsp::delConfig() + 60
```

stack

to replicate

build short circuit
insert it in reaper
close the ui

PR #438, opened 2026-02-23
declaration of audio busses, also better thread handling

defiantnerd:
So, now:

- only inputs can be aux/sidechain inputs
- and actually, all inputs are side chains except the one main bus
- outputs are always standard outputs, since it makes no sense to have sidechain outputs

Also:
- declare main thread semantically on a few calls to the CLAP

### DIFF
--- a/src/wrapasvst3.cpp
+++ b/src/wrapasvst3.cpp
@@ -703,7 +703,7 @@ static Vst::SpeakerArrangement speakerArrFromPortType(const char* port_type)
 void ClapAsVst3::addAudioBusFrom(const clap_audio_port_info_t* info, bool is_input)
 {
   auto spk = speakerArrFromPortType(info->port_type);
-  auto bustype = Vst::BusTypes::kMain;    // actually, everything is main, except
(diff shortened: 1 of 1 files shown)


baconpaul:
@defiantnerd  when I coded this up to test the force fix for S1 I needed two things. The change you have here and another change further down in `getBusInfo` which has to return a result consistent with addAudioBusFrom

Around line 1700 you will need

```
tresult ClapAsVst3::getBusInfo(Vst::MediaType type, Vst::BusDirection dir, int32
         bus.channelCount = info.channel_count;
         bus.direction = dir;
         bus.busType = (info.flags & CLAP_AUDIO_PORT_IS_MAIN) ? Vst::kMain : Vst::kAux;
+        if (dir == Vst::BusDirections::kOutput)
+          bus.busType = Vst::kMain;
+
         bus.flags = Vst::BusInfo::kDefaultActive;
 ```

defiantnerd:
I've added your logic, too, and updated the PR.

baconpaul:
Perfect. merging into next.

PR #439, opened 2026-03-03
Provide an X11 standalone wrapper

baconpaul:
As big libs go to wayland only and stuff, write a small simple X11 poll based wrapper for the standalone. This actually makes it on by default and I think thats appropriate. Later we should remove that GTK based code

### DIFF
--- a/src/detail/standalone/linux/x11_gui.cpp
+++ b/src/detail/standalone/linux/x11_gui.cpp
@@ -1,0 +1,273 @@
+
+
+#include <X11/Xlib.h>
+#include <X11/Xutil.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/timerfd.h>
+#include <unistd.h>
+#include <poll.h>
+#include <string.h>
+#include <cstdint>
+
+#include "x11_gui.h"
+#include <sys/epoll.h>
+
+namespace freeaudio::clap_wrapper::standalone::linux_standalone
+{
+void X11Gui::initialize(freeaudio::clap_wrapper::standalone::StandaloneHost* sah)
+{
+  display = XOpenDisplay(nullptr);
+  sah->x11Gui = this;
+  sah->onRequestResize = [this](int w, int h) { return resetSizeTo(w, h); };
+  epoll_fd = epoll_create1(EPOLL_CLOEXEC);
+  if (epoll_fd < 0)
+  {
+    LOGINFO("Unable to create epoll");
+  }
+}
+
+void X11Gui::runloop()
+{
+  if (!display || window == 0 || epoll_fd < 0)
+  {
+    // no gui. Only way to kill us is with a signal.
+    while (true)
+    {
+      std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+    }
+    return;
+  }
+  XEvent e;
+  struct epoll_event events[maxEpollEvents];
+  bool running{true};
+  while (running)
+  {
+    while (XPending(display))
+    {
+      XNextEvent(display, &e);
+      switch (e.type)
+      {
+        case MapNotify:
+        {
+          clap_window win;
+          win.api = CLAP_WINDOW_API_X11;
+          win.x11 = window;
+          auto ui = plugin->_ext._gui;
+          ui->set_parent(plugin->_plugin, &win);
+          ui->show(plugin->_plugin);
+        }
+        break;
+        case ClientMessage:
+        {
+          // 3. Check if the message is the delete request
+          if ((Atom)(e.xclient.data.l[0]) == wmDeleteMessage)
+          {
+            running = false;
+          }
+          break;
+        }
+      }
+    }
+    // Poll for both X11 events and timer events
+    int num{0};
+    if (running) num = epoll_wait(epoll_fd, events, maxEpollEvents, 50);
+
+    if (num > 0)
+    {
+      for (int i = 0; i < num; ++i)
+      {
+        auto fd = events[i].data.fd;
+        auto tfd = fdToTimerId.find(fd);
+        auto pfd = registeredFds.find(fd);
+
+        if (tfd != fdToTimerId.end())
+        {
+          if (plugin->_ext._timer)
+          {
+            plugin->_ext._timer->on_timer(plugin->_plugin, tfd->second);
+          }
+        }
+        if (pfd != registeredFds.end())
+        {
+          if (plugin->_ext._posixfd)
+          {
+            plugin->_ext._posixfd->on_fd(plugin->_plugin, fd, pfd->second);
+          }
+        }
+      }
+    }
+  }
+}
+
+void X11Gui::setPlugin(std::shared_ptr<Clap::Plugin> p)
+{
+  this->plugin = p;
+  if (display && plugin->_ext._gui)
+  {
+    auto ui = plugin->_ext._gui;
+    auto p = plugin->_plugin;
+    if (!ui->is_api_supported(p, CLAP_WINDOW_API_X11, false))
+    {
+      LOGINFO("[ERROR] CLAP does not support X11");
+      window = 0;
+      return;
+    }
+
+    ui->create(p, CLAP_WINDOW_API_X11, false);
+
+    uint32_t w, h;
+    ui->get_size(p, &w, &h);
+    ui->adjust_size(p, &w, &h);
+
+    int s = DefaultScreen(display);
+    window = XCreateSimpleWindow(display, RootWindow(display, s), 10, 10, w, h, 1,
+                                 BlackPixel(display, s), WhitePixel(display, s));
+    XStoreName(display, window, plugin->_plugin->desc->name);
+    XSelectInput(display, window, InputOutput | StructureNotifyMask);
+
+    // Get window clsoed notifications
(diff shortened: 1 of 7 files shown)


baconpaul:
@defiantnerd I want to re-review this in the morning and maybe consider switching the default for a little before we merge to next.

baconpaul:
OK @defiantnerd I think I would happily include this in .14

defiantnerd:
I agree. I like the minimalism of it.

baconpaul:
yeah and then once we drop .14 we can remove the old gtk stuff altogether. Leave it off-on-a-switch for now.

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+    return;
+  }
+  XEvent e;
+  bool running{true};
+  while (running)
+  {
+    while (XPending(display))
+    {
+      fflush(stdout);
```
baconpaul: Oh this should go

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,208 @@
...
+  this->plugin = p;
+  if (display && plugin->_ext._gui)
+  {
+    auto ui = plugin->_ext._gui;
+    auto p = plugin->_plugin;
+    if (!ui->is_api_supported(p, CLAP_WINDOW_API_X11, false))
+      LOGINFO("[ERROR] CLAP does not support X11");
+
+    ui->create(p, CLAP_WINDOW_API_X11, false);
+    ui->set_scale(p, 1);
```
prokopyl: I don't think this is correct, from my understanding of set_scale at least.

I think it should either use the X11 system scaling factor / DPI setting (since there's nothing else on X11...), or not provide a scale hint at all and explicitly let the plugin figure it out itself, rather than using an incorrect one.

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,208 @@
...
+  *tid = id;
+
+  timerIdToFd[id] = tfd;
+  fdToTimerId[tfd] = id;
+
+  return true;
+}
+bool X11Gui::unregister_timer(clap_id tid)
+{
+  return true;
```
prokopyl, line 217: Looks like the implementation here is missing?

src/detail/standalone/linux/x11_gui.h
```diff
@@ -0,0 +1,47 @@
...
+  void runloop();
+  void shutdown();
+
+  bool register_timer(int period_ms, clap_id* tid);
+  bool unregister_timer(clap_id tid);
+
+  bool register_fd(int fd, clap_posix_fd_flags_t flags);
+  bool unregister_fd(int fd)
+  {
+    return false;
```
prokopyl: Looks like a missing implementation here too

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,208 @@
...
+  int tfd = timerfd_create(CLOCK_MONOTONIC, 0);
+  struct itimerspec ts;
+  memset(&ts, 0, sizeof(ts));
+  ts.it_interval.tv_nsec = period_ms * 1e6;  // Repeat every 1s
+  ts.it_value.tv_nsec = period_ms * 1e6;     // First expiry in 1s
+  timerfd_settime(tfd, 0, &ts, NULL);
+
+  auto fdi = numFds;
+  numFds++;
+  fds[fdi].fd = tfd;
```
prokopyl: `numFds` is not checked, this will lead to buffer overflow if more than 31 FDs + timers are registered by the plugin

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+void X11Gui::setPlugin(std::shared_ptr<Clap::Plugin> p)
+{
+  this->plugin = p;
+  if (display && plugin->_ext._gui)
+  {
+    auto ui = plugin->_ext._gui;
+    auto p = plugin->_plugin;
+    if (!ui->is_api_supported(p, CLAP_WINDOW_API_X11, false))
+      LOGINFO("[ERROR] CLAP does not support X11");
```
prokopyl, line 112: This logs but does not actually return!

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,208 @@
...
+          {
+            running = false;
+          }
+          break;
+        }
+      }
+    }
+    // Poll for both X11 events and timer events
+    int num{0};
+    if (running) num = poll(fds, numFds, 50);
```
prokopyl: I think it would be advantageous to use epoll here to let the kernel handle the FD interest list instead of doing it ourselves.

This would also resolve some of the other issues around fd/timer registration/deregistration I've noted.

Only trade-off is it requires Linux 2.6 & glibc 2.3.2. I don't see any minimum requirements for clap-wrapper, but I think that's fine?

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,208 @@
...
+bool X11Gui::register_fd(int fd, clap_posix_fd_flags_t iflags)
+{
+  int flags{0};
+  if (iflags & CLAP_POSIX_FD_READ) flags = flags | POLLIN;
+  if (iflags & CLAP_POSIX_FD_WRITE) flags = flags | POLLOUT;
+  if (iflags & CLAP_POSIX_FD_ERROR) flags = flags | POLLERR;
+
+  auto fdi = numFds;
+  numFds++;
+  fds[fdi].fd = fd;
```
prokopyl: Same issue here as in `register_timer`

prokopyl (changes requested):

src/detail/standalone/linux/x11_gui.cpp
baconpaul: Yeah there's no real way to get that info reliably in X so I picked '1' in case plugins rely on getting at least one call, but indeed i can kill this.

src/detail/standalone/linux/x11_gui.cpp
baconpaul: epoll makes the code much cleaner yes

src/detail/standalone/linux/x11_gui.cpp
prokopyl: Actually yes you can get this info reliably, it's from the `Xft.dpi` entry in the X resource manager that most Desktop Environments set from the value in system settings.

The downside of this is that this setting is system-wide and not per-monitor, but that is an X11 limitation in general.

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,273 @@
...
+  if (iflags & CLAP_POSIX_FD_READ) flags = flags | EPOLLIN;
+  if (iflags & CLAP_POSIX_FD_WRITE) flags = flags | EPOLLOUT;
+  if (iflags & CLAP_POSIX_FD_ERROR) flags = flags | EPOLLERR;
+
+  epoll_event event;
+  event.events = flags;
+  event.data.fd = fd;
+  if (epoll_ctl(epoll_fd, EPOLL_CTL_ADD, fd, &event) == -1)
+  {
+    LOGINFO("Unable to register display epoll");
```
prokopyl: Nitpick: I think this is a missed copy-paste from above? ^^

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+namespace freeaudio::clap_wrapper::standalone::linux_standalone
+{
+void X11Gui::initialize(freeaudio::clap_wrapper::standalone::StandaloneHost* sah)
+{
+  display = XOpenDisplay(nullptr);
+  sah->x11Gui = this;
+  sah->onRequestResize = [this](int w, int h) { return resetSizeTo(w, h); };
+  epoll_fd = epoll_create1(0);
+  if (epoll_fd < 0)
```
prokopyl, line 24: Nitpick: In general I prefer to use the `EPOLL_CLOEXEC` so that the file descriptor remains private to this process.

It's kind of a nitpick because I don't think clap-wrapper really needs to deal with `fork`s or child processes much if at all, it's just a general habit of mine to keep everything private unless needed. :)

src/detail/standalone/linux/x11_gui.cpp
```diff
@@ -0,0 +1,273 @@
...
+  {
+    XCloseDisplay(display);
+  }
+}
+
+int X11Gui::nextTimerId{2112};
+
+bool X11Gui::register_timer(int period_ms, clap_id* tid)
+{
+  int tfd = timerfd_create(CLOCK_MONOTONIC, 0);
```
prokopyl: Actually same nitpick as with the `epoll_create` call above, this could also use a `TFD_CLOEXEC` flag for the same reasons. (I just missed it on the first review)

prokopyl (changes requested):
Tiny nitpicks, otherwise this looks good to me :)

baconpaul:
OK @defiantnerd this one is good to go once CI passes!

prokopyl (approved):

PR #448, opened 2026-03-07
Make it so reasonable defaults for AUV2 cmake win

baconpaul:
- If you have an auv2 extension and an embedded clap that wins
- If you have cmake args, one plugin, and no auv2 that wins
- If you have cmake args, > 1 plugin, no extension: error
- If you have 1 plugin and auv2, add an option to prefer cmake


baconpaul:
Tested by forcing the embedded example into every configuration; and making sure six sines (multi-out auv2 with extension) was unchanged.

PR #450, opened 2026-03-09
Fix VST3 memory leaks (2)

Quant1um:
Same as #431 but without touching the `_wrappedview` or any other FObject variables:
- remove `addRef()` for `Vst3Parameter`
- use `unique_ptr` for audio buffers in the `ProcessAdapter`

Tested with ASAN/vst3-validator and Bitwig with clapfirst example; if anyone can test this on a plugin with a gui (just in case) that'd be awesome

### DIFF
--- a/src/detail/vst3/parameter.cpp
+++ b/src/detail/vst3/parameter.cpp
@@ -131,9 +131,7 @@ Vst3Parameter *Vst3Parameter::create(
   else
     v.stepCount = 0;
 
-  auto result = new Vst3Parameter(v, info);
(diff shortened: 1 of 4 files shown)


baconpaul:
I can build with short circuit later but short circuit and six sines are both easy to build and have guis if you want test cases in the future

Quant1um:
Just tested with Six Sines locally (Bitwig/Linux), opening/closing the UI does not cause a crash

baconpaul:
Also confirmed short circuit is fine on this branch.

Let me review it now and if it's OK, merge it and move shortcircuit to next. Then let it sit in beta folks hands until the weekend.

Issue #427, opened 2026-01-15
Feature Request: explicit parameter order for AUv2


Quant1um:
It would be nice if there was a way to explicitly specify the order of parameters, or have a mechanism similar to versionHint in JUCE to make it easier to add parameters between releases in hosts that use indices for automation (like Logic Pro) without breaking existing projects.

One possible solution could be: adding a new extension (`clap_plugin_as_auv2`) with a function like `uint32_t get_param_index_for_id(clap_plugin* plugin, clap_id param)` to explicitly specify parameter index for each parameter. Gaps between returned indices could be used to create hidden "tombstone" parameters to preserve the order in case of parameter removal.

Another similar "nice to have" feature would be the ability to specify parameter IDs explicitly for both VST3 and AUv2. A use case for this would be porting a plugin from a different framework to clap-wrapper

baconpaul:
Ahh yeah those are both good ideas.

basically a 'clap param to external param by plugin flavor' extension.

baconpaul:
i'll be facing this problem this quarter when i add a bunch of stuff to six sines.

dathinaios:
Just to say that I got hit by this when I noticed my past Logic saves using my plugin had irrecoverably messed up automation. Took me some time and a bit of detective work to figure out that I have to pay attention to the parameter ID hash to ensure the ordering. So +1 to a way to manage that 🙂

defiantnerd:
an appropriate extensions sounds like a good idea - let me think about it.

baconpaul:
I think the way that juce did it is nice. You have a 'version' tag with your ids and that's the first sort item

Right now we have param by index which is reasonable. But instead param by version then index is the way to go.

baconpaul:
but a better way to do it is simply to have an extension like `bool get_param_order(size_t *order, size_t param_count)` where you just return the indices in order, then leave it up to the plugin. No need to externalize any strategy. On top of that I could do the version thing in my toolset.

defiantnerd:
Is this also a thing for vst3? Since the IDs are definitely the IDs and not the order?

baconpaul:
No this is a logic (and I think live with AU) terrible feature that it stores automation by parameter index not id, like a VST2.

If we had a VST2 wrapper we would have this problem though indeed.

Quant1um:
An extension like
```
typedef struct clap_plugin_params_as_auv2 {
[...]

PR #451, opened 2026-03-14
Implement an extension for auv2 parameter ordering

baconpaul:
In logic and garage band, auv2 parameter ordering in the parameter index list matters for automation, so adding parameters needs a way to set the explicit ordering (as opposed to the default param-id ordering that the AUBase class will use as cast before this commit).

Addresses #427

### DIFF
--- a/include/clapwrapper/auv2.h
+++ b/include/clapwrapper/auv2.h
@@ -35,11 +35,19 @@ extern "C"
   // clap_plugin_auv2_param_ordering extension allows you to provide an ordering for your params
   // by mapping the get_param_info index (0..num-params) to a different ordering.
   //
-  // The default behavior absent this extension is to sort by parameter id, which is clap param id.
+  // The result of this will be used such that `auv2_index = ordering[clap_index]`. So in a five
+  // parameter case, if you want clap index 0 to appear at auv2 position 3, and clap index 3 to appear
+  // at auv2 position 0, you would return `3 1 2 0 4`
+  //
+  // The default behavior absent this extension is to sort by auv2 parameter id, which is clap param id.
   // So if you use a strategy where you go from not adopting this to adopting this when you add params,
   // make sure your old version param subset retains the param id ordering.
   //
-  // For instance, a good impementation if your parameters have a 'version' increasint parameter could be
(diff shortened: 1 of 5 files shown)


baconpaul:
Here's a few things I tried to do carefully

1. If there's no extension, nothing changes
2. Document the extension requirements when you add a parameter with an example
3. Use base classes still for cases we don't handle 

I tested this as follows

took six sines out of the box (no extension) and automated a param with the prior next
added a param - automation broke
wrote this code and didn't implement - automation still broke
implemented the extension to put that param at the end with ordering - automation worked
removed the extension - automation broke
removed the extra param but kept the new code - automation worked

which I think gets all the cases

Quant1um:
Looks good!

One thing we could do is to make the extension return CLAP IDs and not indexes (just how everything else in CLAP uses parameter IDs): one advantage to doing this is that it would simplify the wrapper code significantly as you would be able to just use the ID ordering provided by the plugin in `GetParameterList` directly without doing a conversion step. That is just a suggestion, I think it could work fine as-is.

Oh, and it is probably a good idea to document if it is a `auv2_index = ordering[clap_index]` mapping or a reverse mapping (I can imagine how someone might get confused by that)

defiantnerd (approved):
I've took a deeper look into it and I think it is good.
Also the way a CLAP dev would handle its own order for the plugins.
We should add a little code example for this in the comments of the extension .h file.

baconpaul:
@Quant1um I thought about making it return clap ids but when i had it that way it was actually quite a bit more confusing to debug. Since you can't just print the index ordering or add a sanity checker as easily (and based on some other feedback from @defiantnerd I added a sanity constraint check when we call this). I'll push that up in a second.

Your documentation idea is a good. I'll add that.

@defiantnerd I thought I did include a little bit of code example in the auv2.h and I see it here. Did you want me to change it?

baconpaul:
OK I expanded the documentation in auv2.h quite a bit as well as adding a sanity check in the wrapper. I think this is good to go. If you want another look I'll wait until later in the day to merge but my intent, unless someone screams tonight, is to include this in 0.14

defiantnerd:
Cool! I'll check it out.

defiantnerd:
Happy! I like the limits::max trick!

Go from me!

PR #454, opened 2026-03-31
Statically link libc and libc++ to vst3 when using mingw to build

shockham:
When building a VST3 for Windows on Linux using MinGW the resultant plugin would not work due to the missing libraries as they were dynamically linked.

Not sure if this is the best place or way of fixing this, so welcome to suggestions or updates to the PR.

### DIFF
--- a/cmake/base_sdks.cmake
+++ b/cmake/base_sdks.cmake
@@ -189,6 +189,10 @@ function(guarantee_vst3sdk)
 
 
     if (NOT TARGET vst3_validator)
+        if(MSVC_VST3_VALIDATOR)
+            set(MSVC_VALIDATOR_FLAG -DCMAKE_CXX_COMPILER="MSVC" -DCMAKE_C_COMPILER="MSVC" -G "Visual Studio 17 2022" -A x64)
+        endif()
+
         add_custom_target(vst3_validator)
         add_custom_command(TARGET vst3_validator
                 POST_BUILD
@@ -206,6 +210,8 @@ function(guarantee_vst3sdk)
                         -DSMTG_ENABLE_VSTGUI_SUPPORT=OFF
                         -DSMTG_ENABLE_VST3_PLUGIN_EXAMPLES=OFF
                         -DSMTG_ENABLE_VST3_HOSTING_EXAMPLES=OFF
+
+                        ${MSVC_VALIDATOR_FLAG}
 
                         -B ${CMAKE_BINARY_DIR}/validator-build
 
--- a/cmake/wrap_vst3.cmake
+++ b/cmake/wrap_vst3.cmake
@@ -248,7 +248,8 @@ function(target_add_vst3_wrapper)
                     LIBRARY_OUTPUT_DIRECTORY "${v3root}/${v3root_dor}${V3_OUTPUT_NAME}.vst3/Contents/${v3arch}-win"
                     LIBRARY_OUTPUT_DIRECTORY_DEBUG "${v3root}/${v3root_d}/${V3_OUTPUT_NAME}.vst3/Contents/${v3arch}-win"
                     LIBRARY_OUTPUT_DIRECTORY_RELEASE "${v3root}/${v3root_r}/${V3_OUTPUT_NAME}.vst3/Contents/${v3arch}-win"
-                    SUFFIX ".vst3")
+                    SUFFIX ".vst3"
+                    PREFIX "")
 
             # Copy resource directory, if defined
             if(NOT TCLP_RESOURCE_DIRECTORY STREQUAL "")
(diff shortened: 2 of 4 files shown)


baconpaul:
my primary question is: if using mingw to build on windows, would this also work? I notice we don't have a CI stage for that but I think the GitHub windows runners do provide mingw.

shockham:
Good question, in order to verify am happy to add them to the matrix for the `build_feature_clapfirst-test` following the lead of the `build_feature_cpm_download` job which appears to cater for `gcc/minGW` already.

Is there a reason why the verifier is not set to run for single file windows builds?

Also noticed have added so would only impact the windows folder vst3, so will move to work with the single file build too.

baconpaul:
> Good question, in order to verify am happy to add them to the matrix for the `build_feature_clapfirst-test` following the lead of the `build_feature_cpm_download` job which appears to cater for `gcc/minGW` already.
> 
> Is there a reason why the verifier is not set to run for single file windows builds?
> 
> Also noticed have added so would only impact the windows folder vst3, so will move to work with the single file build too.

The verifier doesnt run for single file windows builds because the 3.8.0 verifier fails on non folder builds as invalid vst3s

cmake/base_sdks.cmake
```diff
@@ -189,6 +189,10 @@ function(guarantee_vst3sdk)
 
 
     if (NOT TARGET vst3_validator)
+        if(MSVC_VST3_VALIDATOR)
+            set(MSVC_VALIDATOR_FLAG -DCMAKE_CXX_COMPILER="MSVC" -DCMAKE_C_COMPILER="MSVC" -G "Visual Studio 17 2022" -A x64)
+        endif()
```
shockham, line 194: Had some trouble getting the validator to build under MinGW, so figured using the MSVC build might be ok here?

cmake/wrap_vst3.cmake
```diff
@@ -248,14 +248,19 @@ function(target_add_vst3_wrapper)
                     LIBRARY_OUTPUT_DIRECTORY "${v3root}/${v3root_dor}${V3_OUTPUT_NAME}.vst3/Contents/${v3arch}-win"
                     LIBRARY_OUTPUT_DIRECTORY_DEBUG "${v3root}/${v3root_d}/${V3_OUTPUT_NAME}.vst3/Contents/${v3arch}-win"
                     LIBRARY_OUTPUT_DIRECTORY_RELEASE "${v3root}/${v3root_r}/${V3_OUTPUT_NAME}.vst3/Contents/${v3arch}-win"
-                    SUFFIX ".vst3")
+                    SUFFIX ".vst3"
+                    PREFIX "")
```
shockham, line 252: The validator wouldn't find the shared lib in the folder directory due to "lib" prefix being applied, so this strips that.

tests/clap-first-example/distortion_clap_entry.cpp
```diff
@@ -17,10 +17,10 @@ extern "C"
 #pragma GCC diagnostic ignored "-Wattributes"  // other peoples errors are outside my scope
 #endif
 
-  const CLAP_EXPORT struct clap_plugin_entry clap_entry = {CLAP_VERSION, dist_entry_init,
+  CLAP_EXPORT extern const struct clap_plugin_entry clap_entry = {CLAP_VERSION, dist_entry_init,
```
baconpaul, line 20: ahh was just about to merge and i saw this change. Why the extern? this is fully defined here right? Sorry to be late seeing ghat.

shockham: No worries! It was to fix this issue when building with MinGW on windows: https://github.com/free-audio/clap-wrapper/actions/runs/23811973893/job/69513675175#step:6:176

For some reason MinGW seems to expect it even though it is already in the [clap entry header](https://github.com/free-audio/clap/blob/main/include/clap/entry.h#L132) and the issue isn't present when using MinGW on linux.

The other compilers appear to be okay with it, but doesn't feel ideal.

baconpaul: hm and does it work? I wonder if the problem is actually that CLAP_EXPORT shoudl not be defined as dllexport on mingw but instead as attribute("export") or some such?

shockham: I believe it works, as passes the VST3 validation, though I have not tested in a DAW myself

baconpaul: OK yeah i'll add it to my weekend todo list when i review all the pre here. I'll either merge it or push it in an #if MINGW of some form.

It looks super weird to declare something as extern which isn't extern is basically my blink reaction.

defiantnerd: I don't think that is right. "extern" is not a declaration for exporting a symbol.

baconpaul: yeah thats what i think too

i wonder if the problem is that CLAP_EXPORT is defined improperly for mingw.

PR #466, opened 2026-05-14
first version of AUv3 on iOS

defiantnerd:
This is now state "works on my machine".


defiantnerd:
this also contains the AUv3 for MacOS branch, I would prefer to merge this one to next.

PR #471, opened 2026-06-13
fix: map analyzer plugins to Fx|Analyzer

### LINKED ISSUE #409: Unable to create VST3 with PlugType::kFxAnalyzer.
Hey all,

Plug-ins with the feature `CLAP_PLUGIN_FEATURE_ANALYZER` do not show in Audacity due to it only allowing VST3s of type `PlugType::kFx` to show in the insert effects list. Bitwig, Live, and Reaper are fine with `PlugType::kAnalyzer` being inserted in their effects lists, but the comment describing `kAnalyzer` in ivstaudioprocessor.h says "not selectable as insert plug-in".

CLAP's documentation is not so specific about if the plug-in can be an insert or not, but it has me wondering if `CLAP_PLUGIN_FEATURE_ANALYZER` would be better mapped to `PlugType::kFxAnalyzer`?

rupayon123:
Fixes #409

This maps `CLAP_PLUGIN_FEATURE_ANALYZER` to `PlugType::kFxAnalyzer` and splits combined VST3 category strings before sorting/deduping, so `Fx|Analyzer` keeps `Fx` as the first token instead of producing duplicate or reversed categories.

Checks:
- `git diff --check`

I did not run a full local C++ build in this environment.

### DIFF
--- a/src/detail/vst3/categories.cpp
+++ b/src/detail/vst3/categories.cpp
@@ -115,13 +115,15 @@ static const struct _translate
 std::string clapCategoriesToVST3(const char *const *clap_categories)
 {
   std::vector<std::string> r;
-  auto appendAttribute = [&r](const char *attribute) {
+  auto appendAttribute = [&r](const char *attribute)
+  {
     std::string value(attribute);
     std::string::size_type start = 0;
     while (start <= value.size())
     {
       auto end = value.find('|', start);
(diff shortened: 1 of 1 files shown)


defiantnerd:
I am closing this. The commit breaks sorting for main categories and also starts sorting alphabetically. I've created a proper fix for this.

Issue #475, opened 2026-07-09
VST3: state not restored in Ableton Live — setComponentState is not implemented


olilarkin:
## Symptom

A clap-wrapper VST3 loses its state when an Ableton Live set is reloaded: after reopening the set, all parameters show (and behave as) their default values, even though the plugin's state was saved correctly.

## Diagnosis

Live's Log.txt shows what happens on set reload:

```
info: VST3: Going to restore: <plugin>
info: VST3: plugin processor successfully loaded: ...
info: Vst3: couldn't initialize controller with processor state of <plugin>: not implemented
info: VST3: couldn't hand processor state of <plugin> to controller: not implemented
info: VST3: Restored: <plugin>
```

Live restores the processor via `IComponent::setState` (which works — `ClapAsVst3::setState` forwards to `clap_plugin_state.load`), and then hands the same stream to the edit controller via `IEditController::setComponentState` to initialize its parameter view. `ClapAsVst3` is a `SingleComponentEffect` and does not override `setComponentState`, so the `EditController` base returns `kNotImplemented`.

The result is that Live's parameter layer keeps the default values from plugin creation: the device panel displays defaults, and Live writes those stale values back over the plugin's freshly restored state — so the restore is effectively undone.

## Why validators don't catch it

Both clap-validator (on the underlying CLAP) and pluginval at strictness 8 (including GUI + state-restoration tests) pass: neither drives the controller through `setComponentState` the way Live does.

## Fix

Implement `setComponentState`: load the handed state into the CLAP plugin (idempotent when `setState` already ran, and covers hosts that call the controller side first), then refresh the wrapper's VST3 parameter cache from the live CLAP values — the same sync `param_rescan(CLAP_PARAM_RESCAN_VALUES)` performs. Verified in Ableton Live 12.4: set reload now restores correctly.

PR follows.

Quant1um:
Wouldn't returning `kResultOk` from `setComponentState` work as well? Some frameworks (like nih-plug) seem to do just that, no need to deserialize the state twice.

olilarkin:
Maybe - we do that in iPlug2

CrushedPixel:
Returning `kResultOk` is the correct fix, but plugins are required to request parameter rescan, as this is where clap-wrapper updates `EditController::parameters`. 

Ableton Live reads parameter values from the edit controller after loading state, while other hosts don't, which is why this mismatch only manifested in Ableton Live.

[...]

PR #476, opened 2026-07-09
VST3: implement setComponentState (fixes state restore in Ableton Live)

olilarkin:
Fixes #475.

Ableton Live restores a set by calling `setState` on the processor, then handing the same stream to the controller via `setComponentState` to initialize its parameter view. As a `SingleComponentEffect` the wrapper inherited `EditController`'s `kNotImplemented`, so Live's parameter layer stayed at default values after a reload and wrote those defaults back over the plugin's restored state (Live Log.txt: `couldn't initialize controller with processor state ... not implemented`).

`setComponentState` now:
- loads the handed state into the CLAP plugin — idempotent when `setState` already ran, and covers hosts that call the controller side first;
- refreshes the VST3 parameter cache from the live CLAP values, the same sync `param_rescan(CLAP_PARAM_RESCAN_VALUES)` performs.

Tested:
- Ableton Live 12.4 (macOS arm64): set reload now restores parameter values and plugin state correctly; device panel matches the plugin.
- pluginval strictness 8 incl. GUI/state-restoration tests: SUCCESS (no regressions; note it passed before this fix too — it doesn't exercise `setComponentState` the way Live does).
- clang-format clean against the repo .clang-format.


olilarkin:
This issue and PR were entirely generated by fable. I'll look a bit more deeply into it next week.

defiantnerd:
thanks. but probably Claude missed the macro redefinition of the function name when being compiled as single component.

olilarkin:
Good thought, but `setComponentState` isn't affected by that macro. The workaround in `vstsinglecomponenteffect.h` is:

```cpp
// work around for the name clash of IComponent::setState and IEditController::setState
#define setState setEditorState
#define getState getEditorState
#include "public.sdk/source/vst/vsteditcontroller.h"
#include "pluginterfaces/vst/ivsteditcontroller.h"
#undef setState
#undef getState
```

It only renames the `setState`/`getState` tokens (the controller's own state methods) during that include — `setComponentState` is a distinct token and keeps its name. Two pieces of evidence the override lands in the right vtable slot:

1. It's declared with the `override` keyword, so a rename or signature mismatch would be a compile error.
2. `nm` on the built plugin shows the expected three symbols side by side:

```
T ClapAsVst3::setComponentState(Steinberg::IBStream*)        <- this PR's override
T ClapAsVst3::setState(Steinberg::IBStream*)                 <- IComponent (processor) state
T Steinberg::Vst::EditController::setEditorState(...)        <- macro-renamed controller state, SDK default
```

The macro *is* relevant to the adjacent gap though: Live's log also prints `couldn't get controller state: not implemented` — that's `IEditController::getState` (= `getEditorState` under the macro), the controller's private state, which remains at the `EditController` default. That's a separate, harmless thing for a single-component wrapper (there's no controller-only state to persist) — probably why the `setComponentState` gap next to it went unnoticed.

Re-confirmed with the patched build in Ableton Live 12.4: state now transfers correctly (duplicating a device carries its state over; that goes through the same processor-state → controller handoff that was returning kNotImplemented).

olilarkin:
(says fable). It might be a Live bug. I'll check next week.

CrushedPixel:
Can confirm this fixes the issue, but might be worth testing what @Quant1um posted in the issue:
https://github.com/free-audio/clap-wrapper/issues/475#issuecomment-4945137221

CrushedPixel:
@defiantnerd I disagree with closing this - we haven't conclusively resolved the issue yet. I am still investigating

defiantnerd:
oops - sorry, that was by mistake.

defiantnerd:
Implemented in #499 because because this one didn't merge anymore.

PR #478, opened 2026-07-09
fixing COM GUID swapper

defiantnerd:


src/wrapasvst3_entry.cpp
```diff
@@ -326,6 +326,16 @@ IPluginFactory *GetPluginFactoryEntryPoint()
       TUID lcid;
       memcpy(&lcid, &g, sizeof(TUID));
 
+#if !COM_COMPATIBLE
+      // apply the same COM-compatibility flip as for the plugin CID above,
```
baconpaul, line 330: When is COM_COMPATIBLE defined? Just want to make sure that we retain same id on mac lin and win still which is what the problem was here.

Issue #303, opened 2024-09-04
VST3+Ableton Live does not load parameters properly


lewloiwc:
I believe the actual problem is on Ableton Live's side rather than with the clap-wrapper, but I'd like to hear everyone's opinions.

After load(), even if we send the current parameter values to the host using try_push() in process(), for some reason Live then sends default values back to CLAP, resulting in all parameters being reset to their default values.

Live sends incorrect values to CLAP only during the first sample of process() after load(), so we've managed to successfully load parameters by ignoring it only at that time.

baconpaul:
Which version of live?

lewloiwc:
Confirmed on 11.3.10 and 12.0

lewloiwc:
When the processes are arranged in chronological order, they look like this:
init()
activate()
start_processing()
save()
load()
save()
deactivate()
activate()
start_processing()
process()
[PluginSyncMainToAudio(plugin, process->out_events);](https://raw.githubusercontent.com/nakst/cdn/main/clap-tutorial-part-4-plugin.cpp) (sends parameter values to the host using try_push())
[c99dist_process_event(plug, hdr);](https://github.com/baconpaul/clap-c99-distortion/blob/c0f7decafb651b521c75b6e25d7dbb6fa2470cd6/src/clap-c99-distortion.c#L373) (for some reason, the default parameter value is sent from the host, which causes issues)

baconpaul:
so basically the first process block after a load has the entire set of params as defaults set into the input queue it sounds like?

@defiantnerd did you test in live much with the wrapper? I have not. but it sounds like we should!

lewloiwc:
That’s exactly right!

lewloiwc:
I think this is a separate issue but it might be related so I'm writing it here.

When using modulators for parameters, after saving and loading the project, incorrect values are saved and I cannot load the intended values. Even just duplicating the relevant track results in the duplicated track saving and loading incorrect values.

Using CLAP_PARAM_IS_MODULATABLE did not improve the situation.

![Image](https://github.com/user-attachments/assets/352b5fcf-1ee4-4e03-b1ef-cf0028d5f606)

<details>

<summary>Full code that includes is_ignore and try_push()</summary>

```c
// This file is here to demonstrate how to wire a CLAP plugin
// You can use it as a starting point, however if you are implementing a C++
// plugin, I'd encourage you to use the C++ glue layer instead:
[...]

PR #491, opened 2026-07-26
macOS: attach the onIdle timer to the main run loop (headless parameter flushes)

rigomapoga:
## Problem

`MacOSHelper::attach()` creates its `CFRunLoopTimer` on the **current thread's** run loop. Hosts may call `attach()` from a worker thread — Ableton Live does when restoring a saved set while the plugin editor has never been opened. A worker thread's run loop never runs, so `onIdle` (which services host-side parameter flushes, clap `on_main_thread` callbacks and timers) never fires.

**Symptom:** plugins restored from a saved Live set stay silent / mis-parameterized until their editor is opened once — the host's parameter layer keeps its instantiation defaults and writes them back into the plugin during the first `process()` blocks; opening the editor wakes the main thread and the deferred flush finally runs. Related to the state-restore family of issues (#475, #303).

## Fix

Attach the timer to `CFRunLoopGetMain()`: the main run loop of a UI host always runs, editor or not. `CFRunLoopTimerInvalidate` in `detach()` is loop-independent, so the removal path is unchanged.

## Verification

- Ableton Live (macOS, Apple Silicon): saved set with never-opened editors now restores and plays correctly headless; before this patch the same set was silent until the editor was opened.
- pluginval (VST3, strictness 8) and auval (aumu + aumf) still pass.

### DIFF
--- a/src/detail/os/macos.mm
+++ b/src/detail/os/macos.mm
@@ -80,7 +80,13 @@ void MacOSHelper::attach(IPlugObject *plugobject)
     _timer =
         CFRunLoopTimerCreate(kCFAllocatorDefault, CFAbsoluteTimeGetCurrent() + (kIntervall * 0.001f),
                              kIntervall * 0.001f, 0, 0, timerCallback, &context);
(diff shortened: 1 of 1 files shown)


defiantnerd:
Thanks, we had that sitting here already.

PR #492, opened 2026-07-26
AUv2: implement MIDIEventList (UMP) input — notes are dropped for MIDI 2.0 hosts

rigomapoga:
## Problem

Hosts that adopt the MIDI 2.0 AU input path deliver notes through `MusicDeviceMIDIEventList` as UMP packets. `AUMIDIBase::MIDIEventList()` returns `kAudio_UnimplementedError` by default and the wrapper has no override (only a disabled `#if 0` sketch in `wrapasauv2.cpp`), so those events are dropped before ever reaching the wrapped CLAP plugin.

**Symptom:** the AU renders, parameters and gain paths are healthy, but no note ever arrives — the instrument is simply silent on event-list hosts. `auval` does not catch this because it still drives the legacy `MusicDeviceMIDIEvent` API.

## Fix

Override `MIDIEventList` in `WrapAsAUV2` and translate channel-voice UMP packets onto the existing `ProcessAdapter::addMIDIEvent` path used by the legacy entry, so both APIs share one code path:

- MIDI 1.0 channel voice (message type 2): passed through directly;
- MIDI 2.0 channel voice (message type 4): note on/off, CC and pitch bend downscaled to 7/14-bit (MIDI 2 velocity 0 is mapped to 1, since it is a real note-on there);
- other packet types are skipped with correct word counts.

## Verification

- auval (aumu + aumf) still passes;
- the wrapped instrument receives notes from both the legacy path (Ableton Live, macOS) and the event-list path.

### DIFF
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -307,8 +307,7 @@ class WrapAsAUV2 : public ausdk::AUBase,
   }
 
 #if AUSDK_MIDI2_AVAILABLE
-  OSStatus MIDIEventList(UInt32 inOffsetSampleFrame,
-                         const struct MIDIEventList *eventList) override
+  OSStatus MIDIEventList(UInt32 inOffsetSampleFrame, const struct MIDIEventList *eventList) override
   {
     // Hosts that speak MIDI 2.0 deliver input as UMP packets through
     // MusicDeviceMIDIEventList (e.g. Logic on recent macOS). AUMIDIBase's
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -306,6 +306,88 @@ class WrapAsAUV2 : public ausdk::AUBase,
     return noErr;  //  HandleMIDIEvent(strippedStatus, channel, inData1, inData2, inOffsetSampleFrame);
   }
 
+#if AUSDK_MIDI2_AVAILABLE
+  OSStatus MIDIEventList(UInt32 inOffsetSampleFrame,
+                         const struct MIDIEventList *eventList) override
+  {
+    // Hosts that speak MIDI 2.0 deliver input as UMP packets through
+    // MusicDeviceMIDIEventList (e.g. Logic on recent macOS). AUMIDIBase's
+    // default returns kAudio_UnimplementedError, so without this override
+    // those notes are dropped before they ever reach the CLAP plugin.
+    // Translate channel-voice packets (MIDI 1.0 mt=2 and MIDI 2.0 mt=4)
+    // onto the same addMIDIEvent path the legacy MIDIEvent entry uses.
+    if (!_processAdapter || !eventList) return noErr;
+    const MIDIEventPacket *packet = &eventList->packet[0];
+    for (UInt32 p = 0; p < eventList->numPackets; ++p)
+    {
+      const UInt32 *w = packet->words;
+      UInt32 i = 0;
+      while (i < packet->wordCount)
+      {
+        const UInt32 word = w[i];
+        const UInt32 mt = word >> 28;
+        UInt32 consumed = 1;
+        switch (mt)
+        {
+          case 0x2:  // MIDI 1.0 channel voice, one word
+          {
+            const UInt32 status = (word >> 16) & 0xFFU;
+            const UInt32 d1 = (word >> 8) & 0x7FU;
+            const UInt32 d2 = word & 0x7FU;
+            _processAdapter->addMIDIEvent(status, d1, d2, (UInt32)packet->timeStamp);
+          }
+          break;
+          case 0x4:  // MIDI 2.0 channel voice, two words
+          {
+            consumed = 2;
+            if (i + 1 >= packet->wordCount) break;
+            const UInt32 w1 = w[i + 1];
+            const UInt32 opcode = (word >> 20) & 0xFU;
+            const UInt32 chan = (word >> 16) & 0xFU;
+            const UInt32 d1 = (word >> 8) & 0x7FU;
+            if (opcode == 0x9)  // note on (16-bit velocity -> 7)
+            {
+              UInt32 v7 = (w1 >> 25) & 0x7FU;
+              if (v7 == 0) v7 = 1;  // MIDI 2 velocity 0 is a real note-on
+              _processAdapter->addMIDIEvent(0x90U | chan, d1, v7, (UInt32)packet->timeStamp);
+            }
+            else if (opcode == 0x8)  // note off
+            {
+              const UInt32 v7 = (w1 >> 25) & 0x7FU;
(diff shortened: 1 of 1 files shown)


baconpaul:
@defiantnerd i kicked off ci here but have t looked. Just calling it to your attention 

@rigomapoga can I make logic send me ump in an au in this path without a midi 2 controller? If so I have some tests to do

rigomapoga:
Good question — you don't need MIDI 2.0 hardware at all. Two ways to exercise this path:

**1. Via Logic (or any contract-following host).** Per Apple's docs, hosts deliver through `MusicDeviceMIDIEventList` when the AU declares `kAudioUnitProperty_AudioUnitMIDIProtocol`; the input source doesn't matter (regions, musical typing, a MIDI 1.0 controller) — everything gets wrapped as UMP, MIDI 1.0-protocol words (mt=2) included. Two caveats from what I measured:

- the wrapper currently does **not** declare `AudioUnitMIDIProtocol`, so contract-following hosts fall back to the legacy `MusicDeviceMIDIEvent` today. Declaring it (e.g. `kMIDIProtocol_1_0`) would be the natural follow-up to this PR — this override is the prerequisite so notes don't vanish the moment a host takes the event-list route (some do regardless of the property).
- data point: instrumented in Ableton Live (macOS 15), the AU input was 53× legacy `MIDIEvent`, 0× `MIDIEventList` — consistent with the property not being declared.

**2. Deterministically, no DAW involved.** A ~90-line host that instantiates the AU and calls `MusicDeviceMIDIEventList` directly with a hand-built MIDI-1UP note-on. Measured on a wrapped instrument: **with this PR** the call returns `noErr` and ~1 s of render gives RMS ≈ 0.27 (sound); **without it** the call returns `kAudio_UnimplementedError` and renders silence. `auval` never exercises this path, which is why the gap is invisible to validation.

<details>
<summary>au_ump_probe.c (build: <code>clang au_ump_probe.c -o au_ump_probe -framework AudioToolbox -framework CoreMIDI -framework CoreFoundation</code>, run: <code>./au_ump_probe aumu SUBT MANU</code>)</summary>

```c
#include <AudioToolbox/AudioToolbox.h>
#include <CoreMIDI/CoreMIDI.h>
#include <math.h>
#include <stdio.h>

static OSType fourcc(const char *s) {
  return ((OSType)s[0] << 24) | ((OSType)s[1] << 16) | ((OSType)s[2] << 8) | (OSType)s[3];
}

int main(int argc, char **argv) {
  if (argc < 4) { fprintf(stderr, "usage: %s TYPE SUBT MANU\n", argv[0]); return 2; }
  AudioComponentDescription d = {0};
  d.componentType = fourcc(argv[1]);
  d.componentSubType = fourcc(argv[2]);
  d.componentManufacturer = fourcc(argv[3]);
  AudioComponent c = AudioComponentFindNext(NULL, &d);
  if (!c) { fprintf(stderr, "component not found\n"); return 1; }
  AudioUnit au;
  if (AudioComponentInstanceNew(c, &au)) return 1;

  AudioStreamBasicDescription f = {0};
  f.mSampleRate = 48000; f.mFormatID = kAudioFormatLinearPCM;
  f.mFormatFlags = kAudioFormatFlagIsFloat | kAudioFormatFlagIsNonInterleaved;
  f.mFramesPerPacket = 1; f.mChannelsPerFrame = 2; f.mBitsPerChannel = 32;
  f.mBytesPerPacket = 4; f.mBytesPerFrame = 4;
  AudioUnitSetProperty(au, kAudioUnitProperty_StreamFormat, kAudioUnitScope_Output, 0, &f, sizeof f);
  UInt32 maxFrames = 512;
  AudioUnitSetProperty(au, kAudioUnitProperty_MaximumFramesPerSlice, kAudioUnitScope_Global, 0,
                       &maxFrames, sizeof maxFrames);
  if (AudioUnitInitialize(au)) return 1;

  struct { MIDIEventList list; UInt32 pad[16]; } storage;
  MIDIEventPacket *pkt = MIDIEventListInit(&storage.list, kMIDIProtocol_1_0);
  const UInt32 noteOn = (2u << 28) | (0x90u << 16) | (60u << 8) | 100u;  // mt=2, 0x90, C3, vel 100
  pkt = MIDIEventListAdd(&storage.list, sizeof storage, pkt, 0, 1, &noteOn);
  OSStatus er = MusicDeviceMIDIEventList(au, 0, &storage.list);
  fprintf(stderr, "MusicDeviceMIDIEventList -> %d%s\n", (int)er,
          er == kAudio_UnimplementedError ? " (UNIMPLEMENTED: notes dropped)" : "");

  float bufL[512], bufR[512];
  AudioBufferList abl; abl.mNumberBuffers = 2;
  abl.mBuffers[0].mNumberChannels = 1; abl.mBuffers[0].mData = bufL;
  abl.mBuffers[1].mNumberChannels = 1; abl.mBuffers[1].mData = bufR;
  double sumsq = 0; long n = 0;
  for (int blk = 0; blk < 94; ++blk) {
    AudioUnitRenderActionFlags fl = 0;
    AudioTimeStamp ts = {0}; ts.mFlags = kAudioTimeStampSampleTimeValid; ts.mSampleTime = blk * 512.0;
    abl.mBuffers[0].mDataByteSize = abl.mBuffers[1].mDataByteSize = sizeof bufL;
    if (AudioUnitRender(au, &fl, &ts, 0, 512, &abl)) return 1;
    for (int i = 0; i < 512; ++i) { sumsq += (double)bufL[i]*bufL[i] + (double)bufR[i]*bufR[i]; n += 2; }
  }
  printf("RMS: %.6f\n", sqrt(sumsq / (double)n));
  return 0;
}
```
</details>

(Also pushed a clang-format fix for the CI failure.)

defiantnerd:
Is this a way to use other people's claude? :) Anyway, if we add MIdI2 to the wrapper we should do it consistently and complete. and also only when MIDI2 is understood from the CLAP side. Translating MUDI 1.0 like here is okay, too, but incomplete.

rigomapoga:
Fair point on scope — agreed that MIDI 2 in the wrapper should be consistent and complete, dialect-aware end to end. What this PR aims to be is the narrower correctness baseline: today any event-list input is dropped entirely, so notes are lost on hosts that take that path.

Concretely I'd split it like this:

this PR: never drop notes. MIDI-1UP (mt=2) passes through to the existing legacy path unchanged; MIDI 2.0 channel voice (mt=4) is downscaled to 7/14-bit only as a fallback.
follow-up: when the CLAP declares CLAP_NOTE_DIALECT_MIDI2 (the wrapper already captures this in _midi_understands_midi2 but never uses it), forward UMP natively as clap_event_midi2 instead of downscaling; declare kAudioUnitProperty_AudioUnitMIDIProtocol accordingly (2.0 when understood, 1.0 otherwise) so contract-following hosts — Logic — actually switch to event-list delivery; keep legacy and event-list entries on one shared path.
If you'd rather have it complete in one PR, I can extend this one instead — happy either way, just tell me which shape you prefer.

defiantnerd:
Superseeded by PR #493, which covers the topic MIDI2 extensively.

PR #493, opened 2026-07-29
AUv2: full MIDI 1.0 + MIDI 2.0 (UMP) in/out, plus minor fixes

defiantnerd:
Overhaul MIDI handling in the AUv2 wrapper so it reaches parity with the VST3 backend for MIDI 1.0 and adds MIDI 2.0 / Universal MIDI Packet support in both directions. All MIDI 2.0 paths are guarded by AUSDK_MIDI2_AVAILABLE.

Dialect handling:
- Replace the single preferred==CLAP test with a real switch over the input port's preferred_dialect, validated against its supported_dialects with a MIDI2->MPE->MIDI->CLAP fallback (chooseInputDialect). Both dialects are threaded through ProcessAdapter::setupProcessing.

MIDI 1.0 (input):
- Poly key pressure -> CLAP_NOTE_EXPRESSION_PRESSURE on the CLAP dialect (raw MIDI otherwise); CC/PC/channel-pressure/pitch-bend stay raw MIDI.
- Implement SysEx input (WrapAsAUV2::SysEx) via a per-cycle owned buffer pool so the borrowed clap_event_midi_sysex_t.buffer stays valid.
- StartNote/StopNote now emit real CLAP notes carrying a note_id (with the key embedded in its low 7 bits, so StopNote needs no lookup table) plus a tuning note-expression for fractional pitch.
- Scope active-note bookkeeping to the CLAP dialect (the note union fields are only valid there).

MIDI 1.0 (output):
- Forward NOTE_EXPRESSION (pressure -> poly aftertouch) and MIDI_SYSEX instead of dropping them; add MIDIOutput::addSysEx.
- Remove the bug that echoed incoming notes straight to the MIDI output.

MIDI 2.0 / UMP:
- Override AUBase::MIDIEventList: walk UMP packets, forwarding MIDI 2.0 channel-voice (MT 0x4) as CLAP_EVENT_MIDI2 and MIDI 1.0 (MT 0x2) through the existing byte path; reassemble SysEx7 (MT 0x3) into CLAP SysEx events.
- Advertise/handle the protocol properties: AudioUnitMIDIProtocol (read, reports MIDI2 only when the plugin prefers it) and HostMIDIProtocol (write).
- Output via kAudioUnitProperty_MIDIOutputEventListCallback: MIDIOutput builds a parallel protocol-1.0 UMP list (incl. SysEx7); Render prefers the EventList block when the host set one, else the legacy MIDIPacketList callback. Raw CLAP_EVENT_MIDI2 output is down-converted to MIDI 1.0. Native MIDI1->MIDI2 output upscaling is intentionally left to the framework's protocol conversion.

Wrapper fixes surfaced by clap-validator-plugins:
- Note-only / zero-audio-port plugins (aumi, or no audio-ports extension) now initialize: PostConstructor presents a placeholder silent output bus, and the plugin's real CLAP audio-port counts are passed to process() separately from the AU element counts (placeholder busses are zero-filled). Fixes Initialize failing with kAudioUnitErr_InvalidElement (-10877).
- Stop scanning the CLAP audio-ports extension while the plugin is active (CLAP AP01 contract violation): the port layout is snapshotted at PostConstructor and ValidFormat/SupportedNumChannels read the cache.

Also remove dead/placeholder code and tidy inconsistent property handling.

Verified: builds -Werror; auval clean (Validator Synth aumu + NoteFX aumi, zero contract violations, Test MIDI PASS); MIDI in/out confirmed in Logic Pro.

### DIFF
--- a/.github/workflows/pullreq.yml
+++ b/.github/workflows/pullreq.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Build project
         run: |
-          cmake -S . -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCLAP_SDK_ROOT=deps/clap -DVST3_SDK_ROOT=deps/vst3sdk -DAAX_SDK_ROOT=deps/aax -DCLAP_WRAPPER_OUTPUT_NAME=testplug
+          cmake -S . -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCLAP_SDK_ROOT=deps/clap -DVST3_SDK_ROOT=deps/vst3sdk -DAAX_SDK_ROOT=deps/aax -DCLAP_WRAPPER_BUILD_AAX=TRUE -DCLAP_WRAPPER_OUTPUT_NAME=testplug
           cmake --build ./build --config Debug
 
       - name: Show Build Results
--- a/src/detail/aax/process.cpp
+++ b/src/detail/aax/process.cpp
@@ -7,6 +7,7 @@
 #include "../shared/midi_translation.h"
 #include <algorithm>
 #include <cmath>
+#include <cstring>
 
 namespace
 {
@@ -97,7 +98,8 @@ void AAXProcessAdapter::setupProcessing(const clap_plugin_t *plugin, double samp
                                         const clap_plugin_audio_ports *ext_audio,
                                         Clap::IAutomation *automation,
                                         std::vector<clap_id> &gesturedparameters,
-                                        ParamChangeQueue &inqueue, uint32_t midiportid, bool preferMIDI)
+                                        ParamChangeQueue &inqueue, uint32_t midiportid, bool preferMIDI,
+                                        uint32_t placeholderInChannels, uint32_t placeholderOutChannels)
 {
   _plugin = plugin;
   _ext_param = ext_param;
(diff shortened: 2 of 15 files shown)


.github/workflows/pullreq.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
       - name: Build project
         run: |
-          cmake -S . -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCLAP_SDK_ROOT=deps/clap -DVST3_SDK_ROOT=deps/vst3sdk -DAAX_SDK_ROOT=deps/aax -DCLAP_WRAPPER_OUTPUT_NAME=testplug
+          cmake -S . -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCLAP_SDK_ROOT=deps/clap -DVST3_SDK_ROOT=deps/vst3sdk -DAAX_SDK_ROOT=deps/aax -DCLAP_WRAPPER_BUILD_AAX=TRUE -DCLAP_WRAPPER_OUTPUT_NAME=testplug
```
baconpaul, line 56: intentional?

src/detail/aax/process.cpp
```diff
@@ -60,7 +98,8 @@ void AAXProcessAdapter::setupProcessing(const clap_plugin_t *plugin, double samp
                                         const clap_plugin_audio_ports *ext_audio,
                                         Clap::IAutomation *automation,
                                         std::vector<clap_id> &gesturedparameters,
-                                        ParamChangeQueue &inqueue, uint32_t midiportid, bool preferMIDI)
+                                        ParamChangeQueue &inqueue, uint32_t midiportid, bool preferMIDI,
+                                        uint32_t placeholderInChannels, uint32_t placeholderOutChannels)
```
baconpaul, line 102: curious why aax is here?

baconpaul (commented):
Curious why AAX and AUv2 mix here?

.github/workflows/pullreq.yml
defiantnerd: Yes, AAX wasn't built in the CI yet. Should we separate it?

defiantnerd: It builds only on Mac and x86-64 Windows, not on ARM Windows. Therefore the check with top_level_default.cmake.

PR #496, opened 2026-07-30
Fix AUv2 channel count handling, memory leak

jwagner:
With the current clap-wrapper next it does not seem to be possible to create AUv2 aufx in mono. SetNumberOfElements resets the channel count in PostConstructor and nothing restores it. Leading to the au failing in auval (and logic) with
```
ERROR: Unit now reports it cannot support default channel configuration (it previously reported that it did): in:2, out:2
ERROR: -10868 IN CALL Cannot Set Input Num Channels:2 when unit says it can
```
This fixes the issue be reinitializing the bus. It should also stops leaking of the name string.

Might be worth renaming add*Bus or refactoring but as is this works for me.


defiantnerd:
Thanks for that.

PR #500, opened 2026-08-06
AUv2: fix a use-after-free reading parameter clump names

baconpaul:
Clumps handed out a const char* into its own std::map and CopyClumpName read through it with no lock held, while setupParameters() cleared that map on the main thread on every param rescan. A host is free to build its parameter tree wherever it likes -- Logic does it asynchronously, on a queue of its own, in response to the very PropertyChanged the rescan just sent -- so for plugins which rescan while audio is running the main thread frees the strings the host is in the middle of reading. It faults in CopyClumpName on the host's tree builder thread.

getClump now answers by value under a mutex, and the map is no longer emptied at all. Clump ids outlive the rescan that minted them: a host keeps the id it read from AudioUnitParameterInfo and asks for its name whenever it likes, so renumbering from zero was making them wrong quite apart from the clearing making them dangerous. addClump() dedupes, so the map converges on the module paths the plugin reports; setupParameters() seeds it, so the ids follow parameter order and are settled before the host is told there is anything to read.

_parametertree had the same window. GetParameterInfo CFRetains a Parameter's CFString on the host's thread while updateInfo CFReleases and recreates it on the main one, and the map was being searched and inserted into at once. A mutex now spans setupParameters and the property getters. The audio thread is deliberately not part of it -- SetParameter and the process adapter read the tree under render and must not block -- so SetBypassEffect and onIdle copy what they need out under the lock and act outside it, which also keeps PropertyChanged's synchronous host listeners off it.

Two smaller faults in the same functions:

- CopyClumpName read a desired length of zero as "the empty string". The SDK clamps kAudioUnitParameterName_Full (-1) to zero on the way in, which is exactly how a host asks for the untruncated name, so every clump came back blank. It also returned noErr with a null CFStringRef when the conversion failed, and the host releases that.

- A parameter name that is not valid UTF-8 leaves Parameter::_cfstring null, and it was CFRetained by the caller and CFReleased in the destructor regardless. It falls back to Latin-1, which accepts any byte.

Assisted-by: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -20,6 +20,9 @@
 #include <iostream>
 #include <memory>
 #include <map>
+#include <mutex>
+#include <optional>
+#include <string>
 
 #include "process.h"
 #include "parameter.h"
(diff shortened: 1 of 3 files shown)


defiantnerd (approved):
Sounds good.

PR #504, opened 2026-08-08
Version 0.16.0

defiantnerd:
Release 0.16.0: merge next into main

- AUv2: stepped and enum parameters now report a proper
  AudioUnitParameterInfo.unit instead of corrupting the flags bitfield, and
  kAudioUnitProperty_BypassEffect now drives and follows the plugin's CLAP
  bypass parameter (#489).
- macOS: the wrapper's timer is now attached to the main run loop rather than
  the current thread's run loop, so it fires reliably (#494).
- WCLAP: a self-contained single-file .wclap.wasm is built by default, and the
  .wclap/ directory bundle plus its distributable .tar.gz are only produced
  when a resource directory is defined (#495).
- AUv2: full MIDI 1.0 and MIDI 2.0 (UMP) input and output with proper dialect
  negotiation and SysEx support, reaching MIDI parity with the VST3 wrapper
  (#493).
- AUv2: buses are reinitialized after SetNumberOfElements resets the channel
  count, fixing mono aufx configurations in auval and Logic and plugging a
  bus-name string leak (#496).
- AUv2: effects with no audio ports get a silent stereo in/out facade
  (placeholder bus, SupportedNumChannels, ValidFormat) so they pass auval
  (#497).
- AUv2: silence buffers are created as cleared buffers so reconfiguration no
  longer feeds uninitialized memory to buses during validation (#498).
- VST3: parameter values are synced from the CLAP plugin after state load so
  the host reads an up-to-date parameter cache, fixing #475 (#499).
- AUv2: fixed a use-after-free when hosts read parameter clump names during a
  parameter rescan — clump names are returned by value under a mutex, clump
  ids stay stable across rescans, and CopyClumpName/non-UTF-8 name handling
  were corrected (#500).
- AUv3: the GUI-created flag is cleared when the plugin's gui->create() fails,
  so no resize or destroy callbacks reach a GUI that was never created, and
  creation can be retried (#501).
- Version bumped to 0.16.0 (32dadee).

### DIFF
--- a/.github/workflows/pullreq.yml
+++ b/.github/workflows/pullreq.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Build project
         run: |
-          cmake -S . -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCLAP_SDK_ROOT=deps/clap -DVST3_SDK_ROOT=deps/vst3sdk -DAAX_SDK_ROOT=deps/aax -DCLAP_WRAPPER_OUTPUT_NAME=testplug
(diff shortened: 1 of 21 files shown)


baconpaul:
Looks great to me. Happy for this to become 0.16

Note I didn't review auv3 and have not tested auv2 midi2 / ump but reviewed rest

PR #155, opened 2023-09-18
ref: drive-by cleanups

0ax1:
Minor drive-by cleanups. No functional changes.

PR #508, opened 2026-08-08
VST3: order category tokens by meaning, map analyzers to Fx|Analyzer

defiantnerd:
VST3: order category tokens by meaning, map analyzers to Fx|Analyzer

clapCategoriesToVST3() once built the VST3 category string in the order
the CLAP descriptor listed its features, and it collected every matching
row of the translation table. Both got lost in 9a729a0 (ref: drive-by
cleanups, #155): the table lookup became a find_if which only ever
applies the first matching row, and the order preserving deduplication
became std::sort + std::unique, which sorts the tokens alphabetically.

Alphabetical is wrong for VST3. Hosts read the first token to file the
plugin and the second one as its sub menu, so an ARA plugin advertised
itself as "Fx|OnlyARA|Reverb" and an external instrument as
"External|Instrument". It also discards the one thing the plugin author
did tell us, namely which sub category they consider the important one.

On top of that a combined attribute like "Instrument|Drum" was treated
as one opaque string and could neither be deduplicated nor ordered
against its own parts, so CLAP_PLUGIN_FEATURE_DRUM_MACHINE ended up as
"Drum|Instrument".

Every attribute is split into its tokens now, each token is kept once in
the order it was first encountered, all matching table rows contribute
again, and a stable sort only moves the two kinds of token that may not
sit in the sub menu slot: a main category (Fx, Instrument) to the front,
a trait rather than a category (OnlyARA, External, Mono, ...) to the end.
Everything in between stays in the order the CLAP features were listed,
so "audio-effect, reverb, chorus" yields "Fx|Reverb|Modulation" while
"audio-effect, chorus, reverb" yields "Fx|Modulation|Reverb" - both
collapsed to the latter before.

CLAP_PLUGIN_FEATURE_ANALYZER maps to kFxAnalyzer instead of kAnalyzer.
The latter is documented as "not selectable as insert plug-in", which
kept CLAP analyzers out of the effect lists of hosts honouring that.

Finally the length check for PClassInfo2::subCategories ignored the
separators it appends, left no room for the terminating zero, and the
trailing pop_back() cut the truncation marker off again - besides being
undefined behaviour when no feature matched at all. Separators are
prepended now and the budget is kSubCategoriesSize - 1, so truncation
drops the categories the author ranked last.

Fixes #409

### DIFF
--- a/src/detail/vst3/categories.cpp
+++ b/src/detail/vst3/categories.cpp
@@ -22,15 +22,21 @@
 
       Fx|Modulation|Flanger
 
-    Also the "Fx" could be added multiple times, duplicates will be automatically removed before creating
-    the VST3 category string without changing the order.
+    A VST3 attribute in the table can be a combination itself (like `Fx|Analyzer`), those are split
+    up into their tokens, so the "Fx" can be added multiple times - duplicates are removed before
+    creating the VST3 category string.
 
     Note that most VST3 hosts will use the first token (like `Fx` or `Instrument`) to locate the plugin
     in a specific location and the second token (like `Synth` as a sub menu category in selection menus).
+    The tokens therefore keep the order of your CLAP features, with two exceptions: a main category
+    (`Fx` or `Instrument`) is always moved to the front and tokens which describe a trait rather than
+    a category (`OnlyARA`, `External`, ...) are always moved to the end, because neither of them may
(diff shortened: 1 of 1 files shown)


baconpaul:
Looks great and the sort logic is clear

PR #511, opened 2026-08-12
Don't release the hosted CLAP from static destruction

defiantnerd:
## The problem

A CLAP is a separate binary that dyld loaded *after* the wrapper, so at process exit it is finalized **first**. Releasing `Clap::Library` from a static destructor therefore calls `clap_entry.deinit()` into a plugin whose own statics are already gone.

This shows up with any hosted CLAP that guards its factory teardown with a `std::mutex`: by the time `~Library()` runs, the mutex has been through `pthread_mutex_destroy`, `lock()` fails with `EINVAL`, and the resulting `std::system_error` escapes a `noexcept` destructor into `std::terminate`.

```
std::mutex::lock -> __throw_system_error      EINVAL
<plugin>::destroyAllInstances
<plugin> factoryDeinit
Clap::Library::~Library                       fsutil.cpp:436
__cxa_finalize_ranges  <-  exit
```

Both validators abort with SIGABRT *after* they have already printed their results, so the failure is easy to miss in a log and impossible to miss in an exit code:

```
Result: 46 tests passed, 1 tests failed
libc++abi: terminating due to uncaught exception of type
           std::__1::system_error: mutex lock failed: Invalid argument
```

## The change

Leak the library at process exit — the OS reclaims it anyway — and keep releasing it for genuine module unloads, where every image involved is still alive.

VST3 gets that for free: `moduleinit.cpp` is already part of `base-sdk-vst3`, so a `Steinberg::ModuleTerminator` runs the release from `DeinitModule()` on a real unload (a plugin rescan, say) and is simply not called at exit. AUv2, AUv3 and AAX have no equivalent hook, so there the library is never destroyed.

`~Library()` is deliberately unchanged — it is still correct for callers that own a `Library` with a bounded lifetime, such as the auv2/auv3 build helpers.

### DIFF
--- a/src/wrapasauv2.cpp
+++ b/src/wrapasauv2.cpp
@@ -10,7 +10,13 @@ extern bool fillAudioUnitCocoaView(AudioUnitCocoaViewInfo *viewInfo, std::shared
 namespace free_audio::auv2_wrapper
 {
 
-Clap::Library _library;  // holds the library with plugins
+// Holds the library with plugins. Intentionally never destroyed: at process
+// exit dyld finalizes the hosted .clap before this binary, so running
+// clap_entry.deinit() from a static destructor reaches into a plugin whose
+// own statics are already gone and aborts the host. AUv2 has no
+// module-unload hook to release it from, so it is leaked and the OS reclaims
+// it. See the same note in wrapasvst3_entry.cpp.
+Clap::Library &_library = *new Clap::Library();
 
 #if 0
 --- 8< ---
--- a/src/wrapasvst3_entry.cpp
+++ b/src/wrapasvst3_entry.cpp
@@ -1,3 +1,7 @@
+// Must precede every include: moduleinit.h below resolves
+// std::numeric_limits<>::max(), which the windows.h `max` macro breaks.
+#define NOMINMAX 1
+
 /*
     CLAP AS VST3 - Entrypoint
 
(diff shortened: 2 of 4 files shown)


src/wrapasvst3_entry.cpp
```diff
@@ -1,3 +1,7 @@
+// Must precede every include: moduleinit.h below resolves
+// std::numeric_limits<>::max(), which the windows.h `max` macro breaks.
+#define NOMINMAX 1
```
baconpaul, line 3: did you mean to add this without a guard?

src/wrapasauv2.cpp
```diff
@@ -10,7 +10,13 @@ extern bool fillAudioUnitCocoaView(AudioUnitCocoaViewInfo *viewInfo, std::shared
...
 {
 
-Clap::Library _library;  // holds the library with plugins
+// Holds the library with plugins. Intentionally never destroyed: at process
+// exit dyld finalizes the hosted .clap before this binary, so running
+// clap_entry.deinit() from a static destructor reaches into a plugin whose
+// own statics are already gone and aborts the host. AUv2 has no
+// module-unload hook to release it from, so it is leaked and the OS reclaims
+// it. See the same note in wrapasvst3_entry.cpp.
+Clap::Library &_library = *new Clap::Library();
```
baconpaul, line 19: Any reason this one isn't static?

defiantnerd: right - an oversight, I will fix that. the define, too

Issue #516, opened 2026-08-16
Standalone ship-quality plan (macOS / Windows / Linux)


baconpaul:
This plan was generated in order for @defiantnerd and @baconpaul to divide and conquer standalone upgrades. It is currently confirmed and reviewed only by Anthropic Fable 5 and is posted here for collaboration.

---

# Standalone Ship-Quality Plan

This is the work plan to take the clap-wrapper standalones (macOS, Windows, Linux)
from "works for development" to "ship to end users". It is based on a full read of
`src/detail/standalone/`, `cmake/wrap_standalone.cmake` and friends, and of the three
shipping consumers (six-sines, two-filters, shortcircuit-xt), all of which include the
standalone in their release installers today.

**How to read this document.** Items are grouped into one cross-cutting section
(shared runtime layer), one build/packaging section, and three per-OS sections. Within
each section items are tagged:

- **Critical** — makes the standalone unfit to ship: crashes, data loss, silent failure,
  or a core feature missing.
- **Cleanup** — real defects and polish gaps that should be fixed but that a careful user
  can live with.
- **Modernize** — architectural or UX upgrades, and decisions to make.

Every item has an ID (`CC-`, `BP-`, `MAC-`, `WIN-`, `LIN-`) so issues/PRs can reference
them. File:line references are against the current `next` branch (v0.16 era). A suggested
execution order is at the end — several per-OS complaints are actually one shared-layer
bug, so the ordering matters.

---

## 1. Cross-cutting: shared standalone layer

Files: `src/detail/standalone/standalone_host.{h,cpp}`, `standalone_host_audio.cpp`,
`standalone_host_midi.cpp`, `entry.cpp`.

### Critical

- [ ] **CC-1: Settings envelope — persist audio/MIDI configuration, not just plugin state.**
  `saveStandaloneAndPluginSettings` (`standalone_host.cpp:320-344`) streams only the raw
  CLAP state blob; the comment at `:322-325` admits the envelope for standalone settings
  doesn't exist. Windows papers over this with its own `settings.json`
  (`windows_standalone.cpp:1333-1404`); macOS and Linux persist nothing. Design one shared,
  versioned settings store (a JSON sidecar next to the state file is fine) holding: audio
  API name, input/output device identified **by name + API** (RtAudio 6 numeric device IDs
  are per-instance handles, not stable across sessions), sample rate, buffer size, input
  mute, selected MIDI input ports, and window geometry. Load it before audio starts, via
[...]

PR #517, opened 2026-08-17
Windows standalone: settings persistence, audio API lifecycle, MIDI selection (#516)

defiantnerd:
Windows half of #516, plus the shared-layer items the Windows criticals sit on.

Scope agreed up front: `WIN-1`..`WIN-7`, plus `CC-1`, `CC-4`/`CC-5`, `CC-10`. Three more
came along because a critical was unfixable without them — `CC-6`, `CC-8` (selection half),
`CC-9`. Everything else in #516 is untouched: no macOS, no Linux, no `BP-*`, no Windows
cleanup or modernize.

None of the three headline Windows complaints turned out to be a Windows bug. Two are
missing shared infrastructure and one is a shared lifecycle bug that Windows is merely the
only frontend able to trigger, being the only one with an API picker. So this is
shared-layer first, thin Windows layer on top — which also leaves macOS and Linux able to
adopt the same code rather than solve it again.

## "Audio settings don't seem to save and restore at all"

They were saved. Startup called `loadSettings()` and then `initializeAudio()`
unconditionally, and `initializeAudio` reset device ids, mute state, sample rate and buffer
size to hardware defaults — so everything just restored was overwritten before it was used.
Only the API name survived, which is exactly the one field commit `5f9c936` had fixed.

Rather than reorder those two calls in the Windows frontend, the store moves to the shared
layer, where the comment in `saveStandaloneAndPluginSettings` has been asking for it since
the beginning. `StandaloneSettings` is a flat UTF-8 `key=value` file written next to the
state blob as `standalone-settings.conf`. No new dependency — the repo carries only `fmt`
and `psl`, and a hand-editable config file is what `LIN-11` wants anyway. Unrecognised keys
survive a load/save round trip, so a newer build's settings are not destroyed by an older
one.

Devices and MIDI ports are stored **by name**. RtAudio 6 device ids are per-instance
enumeration handles; the old code persisted the number, so after a restart it asked for
whichever device happened to land on that id.

Windows splits `initializeAudio` into `applyLoadedAudio` / `applyDefaultAudio` /
`selectAudioApi` and takes exactly one of the first two at startup. (`WIN-1`, `WIN-2`, `CC-1`)

First-run buffer size was `getBufferSizes()[0]` — **16 frames**. It now comes from the
shared default of 256, and 16 is gone from the offered list, since no backend runs there
reliably. `openStream` writes back the size it actually granted, so that is what gets
persisted. (`WIN-4`, `CC-10`)

## "WASAPI to ASIO can crash"

`setAudioApi` was one line: construct the new `RtAudio` over the old one. That constructs
the new backend — probing drivers, and ASIO's probe touches process-global SDK state —
while the outgoing backend's realtime thread is still calling us, then destroys the old
instance by assignment, with no stop, no `running=false` handshake and no lock. The
`isStreamRunning()` guard in `startAudioThreadOnImpl` could not help: by the time it runs,
`rtaDac` already points at the new instance. It now stops, deactivates and destroys the old
instance before building the new one. (`CC-4`)

`activatePlugin`/`deactivatePlugin` never took `processLock`, so an API switch, a device
change or OK in the settings panel could deactivate while `process()` was in flight. The
stop-and-wait handshake now lives inside `deactivatePlugin`, so a frontend cannot forget it;
the hand-rolled copy in the Windows `WM_TIMER` handler is gone. (`CC-5`)

The five combo reads in the settings `WM_COMMAND` handler used `CB_GETCURSEL` directly as a
`std::vector` index. That returns `CB_ERR` (-1) when nothing is selected — an out-of-bounds
read, reachable any time a combo is empty or has just been rebuilt, which is precisely what
switching API does. The sample-rate case was worse: it built its rate list from the *input*
device vector indexed by the *input* combo, so picking a rate on an API with no input
devices — routine for output-only ASIO drivers — indexed an empty vector. (`WIN-3`)

## Under-maintained code

**MIDI selection was theatre.** The listbox handler called `saveSettings()`, but
`saveSettings` wrote no MIDI keys, so nothing was stored. Meanwhile `startMIDIThread` opened
every port on launch while the panel showed nothing selected — the UI stated the opposite of
what was happening. Port selection moves into the shared layer, matched by name; a port that
is unplugged right now is logged and *kept* in the settings so it returns when the device
does. "Bind everything" is tracked separately from "an empty selection", so deselecting
every port is now expressible. The listbox is `LBS_MULTIPLESEL` but was driven with
`LB_SETCURSEL` (documented as having no effect on a multi-select listbox) and `LB_GETCURSEL`
compared against `CB_ERR`, so a restored multi-port selection could not have been displayed
even once it was stored. (`WIN-7`, `CC-8` selection half)

**The UTF converters.** Both kept the source length and the destination length in one
variable: it started as the source length, the sizing call overwrote it with the required
destination length, and the conversion call then received it as the source length.
`toUTF16` silently truncates multi-byte input — `Distörtion äöü` comes back as
`Distörtion ä`, one character lost per non-ASCII character. `toUTF8` is worse: it **reads
past the end of the source buffer**, because the UTF-8 form is longer than the UTF-16 form
for anything non-ASCII. These run on the command line, every device name, every
`FormatMessageW` string, and the settings round trip — so on a non-English Windows they run
constantly. (`WIN-6`)

**Lifetimes.** A null return from `mainCreatePlugin` went straight into the `Plugin`
constructor, which dereferences it on its first line. The `Plugin` struct's own
`shared_ptr<Clap::Plugin>` outlived `mainFinish()`, which runs `entry->deinit()` — so the
plugin object was destroyed after the entry it came from had been deinited, a spec violation
and a use-after-free of the library in the dynamically loaded case. The raw `sah` pointer
dangled from the moment `mainFinish` reset the host, while the settings window could still
be handed messages, and every one of its handlers reads `sah`. (`WIN-5`)

Also `getSampleRates` asked the input device for its rates even when input was unused, and
asked twice (`CC-9`); `startMIDIThread` called `exit(EXIT_FAILURE)` when the first
`RtMidiIn` threw, which on Linux is the routine "ALSA sequencer unavailable" case and made
the GUI app vanish without a word (`CC-6`); `stopMIDIThread` reset the pointers but left the
vector holding them; `stopTimer`'s success test was inverted, logging `GetLastError()`
exactly when `KillTimer` had succeeded.

## Two things not in #516

**The Windows standalone did not build.** C++/WinRT falls back to
`<experimental/coroutine>` under C++17 (`base.h:73-75`), and current MSVC STL hard-errors on
that header. The repo builds the standalone at C++17 by default, so it did not compile with
a current toolchain at all. The only thing pulling winrt in was the settings JSON parser, so
replacing it fixed the build as a side effect. `RuntimeObject.Lib` is no longer linked.

**A shutdown hang.** RtAudio reports device *enumeration* failures through the same error
callback as stream failures, and the frontend puts that in a modal dialog. On a machine with
no audio devices that is one dialog per probe, stacked on the message loop — the app looks
hung and cannot be closed. Confirmed against `next` that this predates the branch.
Enumeration is now silenced for display purposes (never for logging), device ids are checked
before `getDeviceInfo` is asked about them, and a missing input device drops input instead
of failing the whole stream.

## Verified

Built and run on Windows 11, MSVC 14.51 (VS 18), clean configure, Release at the repo's
default C++ standard. The settings panel was driven through real Win32 messages — the same
`CBN_SELCHANGE` notifications the combo boxes send when a user picks an item — rather than
by calling into the code directly.

**Audio API switching (`CC-4`).** Built with `RTAUDIO_API_ASIO=ON` and `RTAUDIO_API_DS=ON`
so the combo offered ASIO, WASAPI and DirectSound, and switched repeatedly with a stream
running and the plugin activated. The lifecycle ordering this PR fixes is sound. **But see
the ASIO section below: this PR does not make WASAPI↔ASIO switching safe, because the
remaining crash is not in clap-wrapper.**

**Device, rate and buffer persistence (`WIN-1`, `WIN-2`).** Through the panel: picked the
non-default output device (SPDIF rather than the default speakers), 192000 Hz, and 512
frames; quit; relaunched. All three came back selected, resolved by name. Sample rates are
re-derived per device and the offered list matched the hardware (4000…192000).

**Graceful degradation.** This box has no capture device. The standalone starts output-only,
logs it, and shows no dialog — where before it asked RtAudio about a device id that names
nothing, which raises an error through the error callback and became a modal box.

Also: unknown settings keys preserved across a round trip; malformed lines and an empty key
dropped without crashing; a device name that no longer exists falls back to the default with
a log line and no dialog; MIDI default binds all, a one-port selection opens exactly that
port, and a selected-but-absent port warns and is *retained*; missing plugin id exits 4 with
a dialog instead of dereferencing null; closing the main window **while the settings window
is open** exits cleanly with code 0, which is the `WIN-5` use-after-free scenario.

The `toUTF16` truncation was checked against the real Win32 API, old algorithm versus new,
rather than reasoned about: `Distörtion äöü` → `Distörtion ä` before, exact after.

## The ASIO crash is upstream in RtAudio, and this PR does not fix it

Retested with two ASIO drivers installed (Steinberg Generic Low Latency, Steinberg
built-in). Selecting ASIO **crashes the process** — `0xC000041D`
(STATUS_FATAL_USER_CALLBACK_EXCEPTION), from an access violation at address `0x0`, i.e. a
call through a null function pointer. Tracked down with a vectored exception handler: the
return address on the stack at the fault is inside `asioglld.dll` +0x8AF2 — the driver
itself is calling a host callback that is null.

The null one is RtAudio's. `RtApiAsio::probeDeviceOpen` sets:

```
asioCallbacks.bufferSwitch         = &bufferSwitch;
asioCallbacks.sampleRateDidChange  = &sampleRateChanged;
asioCallbacks.asioMessage          = &asioMessages;
asioCallbacks.bufferSwitchTimeInfo = NULL;      // <- RtAudio.cpp:3804
```

RtAudio answers `kAsioSupportsTimeInfo` with 0, which by the letter of the ASIO SDK should
mean the driver uses `bufferSwitch`. This driver calls `bufferSwitchTimeInfo` regardless —
a common assumption, since the SDK's own host examples always provide it. Jump to 0, process
gone.

Confirmed by patching the vendored RtAudio to implement `bufferSwitchTimeInfo` as a forward
to `bufferSwitch`:

```c
static ASIOTime *bufferSwitchTimeInfo( ASIOTime *, long index, ASIOBool processNow )
{
  bufferSwitch( index, processNow );
  return 0L;
}
```

With that one change, **15 consecutive switches across ASIO / WASAPI / DirectSound, five of
them selecting ASIO, with a live stream — no crash.** Without it, it dies on the first or
third ASIO selection.

That patch is *not* in this PR: it belongs upstream in RtAudio (6.0.1 is what
`guarantee_rtaudio` pins), or in a deliberate decision to carry a patch or a fork here.
Worth its own issue, and it means "wasapi to asio can crash" in #516 is at least partly not
a clap-wrapper bug at all.

What this PR *does* contribute on that path: the `CC-4` teardown ordering is genuinely
wrong and is fixed; and enumeration is no longer run repeatedly, which matters because every
ASIO `probeDevices()` loads, `ASIOInit()`s and unloads every installed driver — that churn
is what produced `RtApiAsio::probeDeviceOpen: ... (Not enough memory to complete the
request.) creating buffers` on this machine before the caching commit.

Note that on plain `next` this crash does not reproduce, but not because it is fixed there:
`next` never gets far enough to open an ASIO stream at all, because it stalls behind
`getDeviceInfo: deviceId argument not found` dialogs first.

## Not tested, and why

`WM_DEVICECHANGE` — the tail of `WIN-3` — is **not** implemented. Device hot-plug is not
something I can trigger here, and shipping untested audio-restart-on-device-change seemed
worse than shipping nothing.

ASIO is **off by default** in a stock clap-wrapper build — `RTAUDIO_API_ASIO` has to be set
on RtAudio directly, which is `BP-12`'s point about these knobs being undocumented
reach-through. So the crash above needs a deliberate build to reach today.

## Notes for the macOS/Linux side

Nothing here changes macOS or Linux behaviour: neither writes a settings file yet, and an
absent file leaves the previous default path exactly as it was. The shared pieces they can
adopt when convenient are `StandaloneSettings` (`CC-1`, and `LIN-2` only needs
`getStandaloneSettingsPath` implemented for it to work on Linux), the
`openMidiPorts`/`getMidiPortNames` model that `MAC-1` needs, and the activate/deactivate
handshake.

One leftover: `AppDelegate.mm:45-52` still has macOS's copy of the restart lock-dance that
`CC-5` makes redundant. It is correct as written, just no longer necessary — left alone
deliberately to avoid colliding with the macOS work.

### DIFF
--- a/cmake/wrap_standalone.cmake
+++ b/cmake/wrap_standalone.cmake
@@ -71,6 +71,7 @@ function(target_add_standalone_wrapper)
             ${CLAP_WRAPPER_CMAKE_CURRENT_SOURCE_DIR}/src/detail/standalone/standalone_host.cpp
             ${CLAP_WRAPPER_CMAKE_CURRENT_SOURCE_DIR}/src/detail/standalone/standalone_host_audio.cpp
             ${CLAP_WRAPPER_CMAKE_CURRENT_SOURCE_DIR}/src/detail/standalone/standalone_host_midi.cpp
+            ${CLAP_WRAPPER_CMAKE_CURRENT_SOURCE_DIR}/src/detail/standalone/standalone_settings.cpp
             )
     target_link_libraries(${salib}
             PUBLIC
@@ -166,7 +167,9 @@ function(target_add_standalone_wrapper)
             target_compile_definitions(${SA_TARGET} PRIVATE _SILENCE_CLANG_COROUTINE_MESSAGE)
         endif()
 
-        target_link_libraries(${SA_TARGET} PRIVATE base-sdk-wil ComCtl32.Lib RuntimeObject.Lib)
+        # RuntimeObject was only ever needed for the C++/WinRT JSON parser the
+        # settings code used to use; the shared settings store replaced it.
+        target_link_libraries(${SA_TARGET} PRIVATE base-sdk-wil ComCtl32.Lib)
 
     elseif(UNIX)
         target_sources(${SA_TARGET} PRIVATE
--- a/src/detail/standalone/standalone_host.cpp
+++ b/src/detail/standalone/standalone_host.cpp
@@ -451,26 +451,35 @@ bool StandaloneHost::isKnownDevice(unsigned int deviceID)
 {
   if (!rtaDac) return false;
 
-  SilenceAudioErrors silence(this);
-
-  for (auto id : rtaDac->getDeviceIds())
+  for (const auto &device : deviceList())
   {
-    if (id == deviceID) return true;
+    if (device.ID == deviceID) return true;
   }
 
   return false;
 }
 
+RtAudio::DeviceInfo StandaloneHost::deviceInfoFor(unsigned int deviceID)
+{
+  for (const auto &device : deviceList())
+  {
+    if (device.ID == deviceID) return device;
+  }
+
+  return {};
+}
+
 std::string StandaloneHost::deviceName(unsigned int deviceID)
 {
-  if (!isKnownDevice(deviceID)) return {};
+  // Answer from the cached enumeration rather than getDeviceInfo(): an id RtAudio
+  // does not recognise raises an error through the error callback, which a
+  // frontend may well be putting in front of the user as a modal dialog.
+  for (const auto &device : deviceList())
+  {
+    if (device.ID == deviceID) return device.name;
+  }
 
-  SilenceAudioErrors silence(this);
-
-  // Only ask once we know the id is real. getDeviceInfo() on an unknown id makes
-  // RtAudio raise an error through the error callback, which a frontend may well
-  // be putting in front of the user as a modal dialog.
-  return rtaDac->getDeviceInfo(deviceID).name;
+  return {};
 }
 
 unsigned int StandaloneHost::resolveInputDevice(const std::string &name)
--- a/src/detail/standalone/standalone_host.cpp
+++ b/src/detail/standalone/standalone_host.cpp
@@ -366,6 +366,147 @@ bool StandaloneHost::tryLoadStandaloneAndPluginSettings(const fs::path &fromDir,
   return true;
 }
 
+std::optional<fs::path> StandaloneHost::standaloneSettingsFile()
+{
+  if (!clapPlugin || !clapPlugin->_plugin || !clapPlugin->_plugin->desc) return std::nullopt;
+
+  auto base = getStandaloneSettingsPath();
+  if (!base.has_value()) return std::nullopt;
+
+  return *base / clapPlugin->_plugin->desc->id / "standalone-settings.conf";
+}
+
+bool StandaloneHost::loadStandaloneSettings()
+{
+  auto path = standaloneSettingsFile();
+  if (!path.has_value()) return false;
+
+  settingsLoaded = settings.load(*path);
+
+  return settingsLoaded;
+}
+
+bool StandaloneHost::saveStandaloneSettings()
+{
+  auto path = standaloneSettingsFile();
+  if (!path.has_value()) return false;
+
+  // The plugin-id directory is shared with the state blobs and usually exists by
+  // now, but a first run which never saved state has yet to create it.
+  try
+  {
+    fs::create_directories(path->parent_path());
+  }
+  catch (const fs::filesystem_error &e)
+  {
+    LOGINFO("[ERROR] Unable to create settings directory '{}': '{}'", path->parent_path().u8string(),
+            e.what());
+    return false;
+  }
+
+  return settings.save(*path);
+}
+
+void StandaloneHost::captureAudioSettings()
+{
+  settings.audioApiName = audioApiName;
+  settings.outputDeviceName = deviceName(audioOutputDeviceID);
+  settings.inputDeviceName = deviceName(audioInputDeviceID);
+  settings.audioOutputUsed = audioOutputUsed;
+  settings.audioInputUsed = audioInputUsed;
+  settings.sampleRate = currentSampleRate;
+  settings.bufferSize = currentBufferSize;
+}
+
+void StandaloneHost::applyAudioSettings()
+{
+  // The API has to be selected first: every device id below is an index into a
+  // particular RtAudio instance's enumeration, so that instance has to be the one
+  // we are about to open a stream on.
+  auto api = RtAudio::Api::UNSPECIFIED;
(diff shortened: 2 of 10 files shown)


baconpaul:
I think if we merge this to next it would be great. Then I can start a VM on my mac cranking linux today at least.

PR #520, opened 2026-08-20
Fix issue519

defiantnerd:


singularpoint:
One small thing: with `-_bootstrapAudioUnit` in place the audio unit can now be created from two threads — the host's XPC worker and the main thread — and the `if (self.audioUnit)` guard doesn't make that safe by itself. If the two ever overlap you get two plugin instances, with no guarantee which one the host keeps.

I never saw them overlap; the ordering just isn't guaranteed anywhere. A `@synchronized` around the factory body would settle it.

defiantnerd:
fix confirmed by @singularpoint

Issue #521, opened 2026-08-21
AUv2: restartPlugin() is a no-op, so request_restart never happens


baconpaul:
### Summary

`WrapAsAUV2::restartPlugin()` is an empty function body, so a hosted CLAP's call to `clap_host::request_restart` does nothing in AUv2. Anything the plugin defers until the next `deactivate()`/`activate()` cycle — latency changes, block-size-dependent reallocation, FFT size, oversampling factor — never takes effect for as long as the AU stays initialized.

Line references are against `be1a04c` (current `main`).

```cpp
// src/detail/auv2/auv2_base_classes.h:583
void restartPlugin() override
{
}
```

### Why it matters

`request_restart` is the only mechanism CLAP gives a plugin for "I need to reallocate; please take me through deactivate/activate". A plugin that follows the spec — does not allocate on the audio thread, defers the change, asks for a restart — silently never applies the change under AUv2.

We hit this in SpectrumWorx: its FFT size and overlap factor reallocate the whole spectral working set, so they are deferred and applied in `clap_plugin::deactivate()`. In CLAP hosts and VST3 the control works; in Logic (AUv2) the value moves in the UI and the audio engine keeps running the old size. Reported downstream as surge-synthesizer/SpectrumWorx#141.

The same plugin's reported latency is also frozen, for the same reason — `clap_plugin_latency::get` only changes across a restart, so `latency_changed()` (which *is* implemented, at `src/wrapasauv2.cpp:1060`) is never reached.

### Current state per format

| format | `restartPlugin()` | |
|---|---|---|
| VST3 | `src/wrapasvst3.cpp:1352` sets `_requestRestart`, consumed at `:1509` → `restartComponent(kIoChanged \| kLatencyChanged)` | ✅ |
| standalone | `src/detail/standalone/standalone_host.h:114` sets `restartRequested`, consumed at `src/detail/standalone/macos/AppDelegate.mm:38` and `windows/windows_standalone.cpp:1006` | ✅ |
| **AUv2** | `src/detail/auv2/auv2_base_classes.h:583` — empty | ❌ |
| AUv3 | `src/detail/auv3/auv3_audiounit.mm:188` — logs only | ❌ |
| AAX | `src/detail/aax/wrapper.cpp:1107` — empty | ❌ |

It has been empty since the AUv2 wrapper landed (`3074ce0`, "Initial Work - AUv2 (#178)").

### Why the AU host doesn't need to be involved

[...]

PR #524, opened 2026-08-21
AUv2: honour request_restart

baconpaul:
A plugin that defers reallocation until its next activate cycle -- a changed FFT size, latency or oversampling factor -- never applied it under AUv2, because restartPlugin() was an empty body. Reported against SpectrumWorx, where the FFT size control moved in the UI while the audio engine went on running the old size in Logic. See #521.

The AU host need not be involved. The wrapper owns the CLAP's activate/deactivate itself, so onIdle can cycle the plugin underneath a still initialized AU, which is what the standalone already does. Render takes the process lock for its whole body so the rebuild can fence against a render already in flight, and holds it only long enough to close the door rather than across the rebuild.

Releasing the host's MIDI output callbacks moves out of deactivateCLAP and into Cleanup. The host hands those over through SetProperty and does not reinstall them across a restart it did not perform, so a plugin with MIDI output would have lost it for the rest of the session.

Assisted-by: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -27,6 +27,7 @@
 #include "process.h"
 #include "parameter.h"
 #include "detail/shared/fixedqueue.h"
+#include "detail/shared/spinlock.h"
 #include "detail/shared/midi_translation.h"
 #include "detail/os/osutil.h"
 #include "detail/clap/automation.h"
@@ -583,6 +584,7 @@ class WrapAsAUV2 : public ausdk::AUBase,
   }
   void restartPlugin() override
   {
+    _requestRestart = true;
   }
   void request_callback() override
   {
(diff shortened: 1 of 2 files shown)


baconpaul:
@defiantnerd mind taking a look here? This fixes a live problem we found day one of SpectrumWorx wider avail in logic. It fixes it and looks right to me but would like your eyes before we press merge.

defiantnerd:
Yes, this was indeed a problem. I didn't implement restart (and even denied it) to the clap-plugin since AU does not allow this. But this would allow the clap to restart and just handle it in the wrapper.
The spinlock is the right choice to protect rendering.

baconpaul:
Yeah this only came u- because spectrumworx can change latency so has to do it with stop start. And so au didn’t allow fft window changes. Thanks!

Issue #437, opened 2026-02-21
Be Better about Busses


baconpaul:
Restating #348 which got accidentally hijacked for something different, which we thought was the same, but wasnt

Over in https://github.com/free-audio/clap-wrapper/pull/346 we added code so we don't respond to a bus configuration which doesn't match the audio ports configuration. This at least means multi out vst3 in the wrapper work in reaper and bitwig

But it's wildly incomplete in two ways

There are extensions we could consult other than audio ports, like configurable-audio ports or audio port configuration which ever is the one which is appropriate, and audio port activation
We could provide wrapper options (perhaps with another extension) to allow things like mono-in-spreads-to-stereo-busses and stereo-out-averages-to-mono-busses for simple effects which don't implement the above. Maybe
So still work to do here for sure.

defiantnerd:
This matches https://github.com/free-audio/clap-wrapper/pull/438 and is now in next

defiantnerd:
This now in main with 0.14

Issue #348, opened 2025-01-22
Be Better about Busses


baconpaul:
Over in #346 we added code so we don't respond to a bus configuration which doesn't match the audio ports configuration. This at least means multi out vst3 in the wrapper work in reaper and bitwig

But it's wildly incomplete in two ways

1. There are extensions we could consult other than audio ports, like configurable-audio ports or audio port configuration which ever is the one which is appropriate, and audio port activation
2. We could provide wrapper options (perhaps with another extension) to allow things like mono-in-spreads-to-stereo-busses and stereo-out-averages-to-mono-busses for simple effects which don't implement the above. Maybe

So still work to do here for sure.

JuliusLC:
I think this issue has something to do with my problem with Studio One not enabling outputs on VST3 Shortcircuit, the same happens with the Clap version.
Evildragon explained me that Presonus/Fender Studio One has an inxomplete Clap spec... But I never had that problem with other VST3 plug-ins & Outputs inside the DAW.
Or maybe it's the time to sport a FST/RST to CLAP adapter to avoid VST3 weird magic.

Maybe a fixed Minimum of 16/32 outs would be better than losing multiouts.

baconpaul:
for shortcircuit, its even easier - there's only one fixed configuration. If the vst3 isn't showing those busses we need to fix the wrapper.

I thought I had debugged this with six sines seven outs though, but lemme look again. 

I'll open an issue in SC repo to check it out.

baconpaul:
OK with shortcircuit xt latest 

- in reaper as VST3 I get prompted "add 2 tracks or 21" and both work as expdcted
- in bitwig as VST3 i can go to the device and add the chains and works as expected
- In logic as AUV2 I can add the 21 stereo out version and stack my aux outs and works as expected

So @JuliusLC I'm not sure what you are seeing in S1 / Presonus. I should grab latest since I think i have an NFR (and anyway just purchased a faderport 8 for my studio) but the wrapper seems to be doing the reasonable thing for the vst3 here

This issue was really about the *other* ways you can specify alternate topologies in a clap. But would love to understand the issue you saw.

JuliusLC:
The fact is that I can see the multiple outputs in Studio One, but they are disabled. there's no sound and i can't check them to be enabled
<img width="838" height="1058" alt="Image" src="https://github.com/user-attachments/assets/af4b08e8-c9bb-4514-b922-07550d3cae03" />
[...]

PR #530, opened 2026-08-28
AUv2: support alternate bus layouts via configurable-audio-ports

defiantnerd:
Addresses the AUv2 half of #348 / #437: the AUv2 wrapper snapshotted the CLAP audio ports once at `PostConstructor` and pinned every bus to that channel count, so a plugin offering several bus layouts was stuck with its startup layout in every AUv2 host (Logic would only offer the one configuration).

This consumes `clap.configurable-audio-ports` the way the VST3 wrapper already does in `setBusArrangements` — and only that extension. AUv2 has no push negotiation, so the wrapper probes:

- **PostConstructor** probes `can_apply_configuration` for main-bus channel pairs (up to 8×8; non-main ports keep their current counts in every request) and snapshots the accepted pairs — the plugin is guaranteed deactivated there, which is the state the extension requires, and like the existing port caches the result is never queried live afterwards.
- **SupportedNumChannels** advertises one `AUChannelInfo` entry per accepted pair (falling back to the existing cross-product when the plugin has no `configurable-audio-ports`).
- **ValidFormat** accepts, on a main bus, any channel count an accepted pair offers on that scope — still validated against the snapshots, preserving the active-state constraint documented there.
- **activateCLAP** gains `applyConfigurationFromBusFormats()`: when the host-chosen main-bus formats differ from the current port layout, it pushes the matching configuration through `apply_configuration` (request shape mirrors `wrapasvst3.cpp`), then refreshes the port caches and bus names from the rescanned ports. Main thread, plugin deactivated — exactly the state the extension requires.

The render path needs no changes: the AUv2 `ProcessAdapter` already sizes and fills its buffers from the AU stream formats rather than the port infos.

Tested with a CLAP whose `can_apply_configuration` accepts mono→stereo (startup), stereo→stereo and mono→mono. `auval -v aumf` before: `Reported Channel Capabilities (explicit): [1, 2]`, only 1-2 rendered. After:

```
[clap-wrapper] PostConstructor: 3 probed main-bus channel layouts

Reported Channel Capabilities (explicit):
      [1, 1]  [1, 2]  [2, 2]

Input/Output Channel Handling:
1-1   1-2   1-4   1-5   1-6   1-7   1-8   2-2   ...
X     X                                   X
```

with the reconfiguration visibly applied for each format the matrix exercises (`applied audio port configuration for 1/1 channels`, `... 2/2 ...`) and `AU VALIDATION SUCCEEDED` throughout. Plugins without `configurable-audio-ports` are unaffected (empty probe cache short-circuits every new path).

### DIFF
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -764,10 +764,15 @@ class WrapAsAUV2 : public ausdk::AUBase,
   // CLAP port layout (ValidFormat admits every probed layout), push the
   // matching configuration into the plugin via configurable-audio-ports and
   // refresh the port caches. Called from activateCLAP: main thread, plugin
-  // deactivated — the state the extension requires.
-  void applyConfigurationFromBusFormats();
+  // deactivated — the state the extension requires. Returns false when the
+  // plugin rejects the configuration, in which case activation must not
+  // proceed (the AU formats and the plugin's ports would disagree about
+  // buffer sizes).
+  bool applyConfigurationFromBusFormats();
 
-  void activateCLAP();
+  // Returns false when the plugin rejects the host-chosen bus formats; the
+  // caller must treat that as a failed initialization.
+  bool activateCLAP();
   void deactivateCLAP();
   // the AU-level half of the teardown, which an internal restart must not do
   void releaseHostMIDIOutput();
@@ -816,7 +821,12 @@ class WrapAsAUV2 : public ausdk::AUBase,
   };
   std::vector<ChannelCapsCache> _channelCapsCache;
 
-  // Upper bound of the per-scope channel counts probed for the cache.
+  // Upper bound of the per-scope main-bus *channel counts* probed for the
+  // cache (not a bus count). The active layout is seeded into the cache
+  // unconditionally, so a plugin whose main bus is wider than this still
+  // advertises its own configuration; the bound only limits how many
+  // alternate layouts are asked about (and so how many
+  // can_apply_configuration calls a single instantiation makes).
   static constexpr uint32_t kMaxProbedChannels = 8;
 
   uint32_t _midi_preferred_dialect = 0;
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -754,6 +754,19 @@ class WrapAsAUV2 : public ausdk::AUBase,
   void addInputBus(int bus, const clap_audio_port_info_t *info);
   void addOutputBus(int bus, const clap_audio_port_info_t *info);
 
+  // Configuration requests for the main busses: main ports get the given
+  // channel counts, non-main ports keep their current ones. Shared by the
+  // PostConstructor probe and applyConfigurationFromBusFormats.
+  std::vector<clap_audio_port_configuration_request_t> mainBusConfigurationRequests(
+      uint32_t mainInChannels, uint32_t mainOutChannels) const;
+
+  // If the host chose main-bus stream formats that differ from the current
+  // CLAP port layout (ValidFormat admits every probed layout), push the
+  // matching configuration into the plugin via configurable-audio-ports and
+  // refresh the port caches. Called from activateCLAP: main thread, plugin
+  // deactivated — the state the extension requires.
+  void applyConfigurationFromBusFormats();
+
   void activateCLAP();
   void deactivateCLAP();
   // the AU-level half of the teardown, which an internal restart must not do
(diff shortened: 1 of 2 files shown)


src/detail/auv2/auv2_base_classes.h
```diff
@@ -790,6 +803,22 @@ class WrapAsAUV2 : public ausdk::AUBase,
...
+  // once the host has settled on formats (applyConfigurationFromBusFormats).
+  struct ChannelCapsCache
+  {
+    uint32_t inputChannels;   // main input channel count, 0 = no main input
+    uint32_t outputChannels;  // main output channel count, 0 = no main output
+  };
+  std::vector<ChannelCapsCache> _channelCapsCache;
+
+  // Upper bound of the per-scope channel counts probed for the cache.
+  static constexpr uint32_t kMaxProbedChannels = 8;
```
baconpaul, line 830: Shortcircuit has 21 outputs. Is this bus or channel count?

src/wrapasauv2.cpp
```diff
@@ -1595,7 +1701,24 @@ bool WrapAsAUV2::ValidFormat(AudioUnitScope inScope, AudioUnitElement inElement,
...
+  // (applyConfigurationFromBusFormats), before it activates.
+  if (cache[inElement].isMain)
+  {
+    const bool isInput = (inScope == kAudioUnitScope_Input);
+    for (const auto &caps : _channelCapsCache)
+    {
+      const uint32_t channels = isInput ? caps.inputChannels : caps.outputChannels;
+      if (channels != 0 && channels == inNewFormat.mChannelsPerFrame) return true;
+    }
+  }
```
baconpaul, line 1768: No else needed here? What about  plugin with only a non main input (like a sidechain only synth)

src/wrapasauv2.cpp
```diff
@@ -1711,6 +1846,52 @@ void WrapAsAUV2::PostConstructor()
...
+  // deactivated here, which is the state can_apply_configuration requires;
+  // like the port scan above, the result is snapshotted and never queried
+  // live afterwards. Every accepted pair is advertised through
+  // SupportedNumChannels, accepted by ValidFormat, and applied in
+  // activateCLAP once the host has settled on formats.
+  if (_plugin->_ext._configurable_audio_ports && (!_inputPortCache.empty() || !_outputPortCache.empty()))
+  {
+    uint32_t currentIn = 0, currentOut = 0;
+    for (const auto &port : _inputPortCache)
+      if (port.isMain)
```
baconpaul, line 1914: Again the inputs with no main case

src/detail/auv2/auv2_base_classes.h
baconpaul: Reading more I’m
Not even sure this is needed at all

baconpaul (commented):
I would need to run this to answer my own question but had some concerns about the branches where you have inputs but none of them are main (which is the synth-with-a-sidechain chase)

defiantnerd:
good point, I will read the AU docs tomorrow how that'll be translated.

src/detail/auv2/auv2_base_classes.h
defiantnerd: Channel count of the main bus, not bus count. And as of 7671b8a the active layout is seeded into the cache before the grid runs, so a plugin whose main bus is wider than 8 (Shortcircuit's 21) still advertises its own configuration — the cap only bounds which *alternates* get asked about.

src/wrapasauv2.cpp
defiantnerd: Fixed in 7671b8a: the probe — and with it this widening path — now only engages when every scope that has ports also has a main port. A sidechain-only input keeps the pre-probe behaviour: the caps cache stays empty, SupportedNumChannels falls back to the port-derived cross-product, and this branch pins every bus to its port's count. No else needed since the final return false covers the miss.

src/wrapasauv2.cpp
defiantnerd: Same fix: with no main port on a populated scope the probe can neither vary that scope nor describe it in an AUChannelInfo pair, so 7671b8a skips it entirely and falls back to the previous advertisement — instead of publishing 0 channels for a scope that does take audio (or, in the nobody-sets-IS_MAIN case, collapsing the whole cache to a single {0,0}).

src/detail/auv2/auv2_base_classes.h
defiantnerd: The grid is standing in for an enumeration the extension doesn't offer: configurable-audio-ports only answers can_apply_configuration pair by pair, and SupportedNumChannels needs concrete pairs up front (scanning the ports live is illegal once the plugin is active). So it's at most 64 main-thread calls, once, at PostConstructor. The alternative would be answering ValidFormat optimistically and only failing at Initialize when apply_configuration refuses — happy to rework it that way instead if you'd prefer less probing.

defiantnerd:
Pushed 7671b8a addressing the review: the probe is skipped when a populated scope has no main port (falls back to the pre-existing cross-product advertisement — this covers the sidechain-only-synth chase from the review), the active layout is seeded ahead of the probe grid so kMaxProbedChannels can't hide a wide plugin's own configuration, and a failed apply_configuration at activation now fails Initialize with kAudioUnitErr_FormatNotSupported instead of rendering with AU buffers sized differently from the plugin's ports.

src/detail/auv2/auv2_base_classes.h
baconpaul: Gotcha so 21 stereo ports is fine but 1 40 channel port may get skipped 

So is 8 wide enough for all the various ambisonic configs?

defiantnerd: ambisonics can have up to several hundred channels, but I would try to limit it to a sane amount until it is needed and we have a real world example.

baconpaul:
OK I'm happy for this to get merged. Once it is, I'll update spectrumworx to implement configurable audio ports and see if I can make a mono-mono version work in logic.

PR #531, opened 2026-08-29
Standalone: honour the plugin's per-bus channel counts in clapProcess

defiantnerd:
The standalone host hard-asserted two channels on every plugin bus (`// For now assert sterep`) and always handed the plugin stereo buffers. Any plugin whose main input or output bus is not stereo — e.g. a mono-input guitar amp — aborts a debug build on the first process call, and a release build lies to the plugin about its channel count.

`clapProcess` now:

- sizes each `clap_audio_buffer` from `inputChannelByBus` / `outputChannelByBus` instead of assuming 2 channels per bus,
- locates the main busses via `mainInput` / `mainOutput` (previously: first bus wins),
- maps between the interleaved device stream and the main bus with a repeat-last/drop-surplus channel policy: a mono device feeds both channels of a stereo bus, a mono bus plays on both channels of a stereo device, and surplus channels on either side are dropped,
- interleaves with the actual device channel counts (`currentInputChannels` / `currentOutputChannels`) instead of a hardcoded stereo stride, so a device that only granted one output channel no longer gets written past its buffer,
- renders silence when there is no main output bus instead of indexing `utilityBuffer` with `-1`.

Three adjacent holes closed along the way (found reviewing the first version of this PR):

- `process.audio_outputs_count` was still pinned to `1` two lines after being set from `numAudioOutputs`, so a multi-bus plugin was only ever told about its first output bus — even though every bus now gets valid buffers, a spec-conforming plugin with its main output on bus 1 would never have written it.
- The device-output writes are now guarded against a null output buffer, and `startAudioThreadOnImpl` resets `currentInputChannels` / `currentOutputChannels` before opening a stream. Previously, reconfiguring from an output-capable stream to an input-only one left the old stream's channel count paired with a null RtAudio output buffer — a crash on the audio thread.
- `setupAudioBusses` clamps the reported busses and channels to the fixed process pools (`utilityBufferMaxChannels` of each) instead of relying on an assert that release builds compile out (a plugin with >64 total channels would have overrun `utilityBuffer` and the stack arrays in `clapProcess`), and resets its accumulators so a rescan cannot double-count.

Verified with a host-side harness driving `clapProcess` with a stub plugin over eleven layouts: mono→stereo, stereo→stereo (unchanged behaviour), mono device in, mono device out, no input device, two output busses with main on bus 1, null device output with a stale channel count, no output busses at all, and the `setupAudioBusses` clamp/rescan cases.

### DIFF
--- a/src/detail/standalone/standalone_host.cpp
+++ b/src/detail/standalone/standalone_host.cpp
@@ -1,4 +1,5 @@
 
+#include <algorithm>
 #include <cassert>
 #include "standalone_host.h"
 #include <fstream>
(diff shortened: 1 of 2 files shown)


baconpaul:
Yeah this looks correct to me indeed. Happy for you to merge it.

PR #533, opened 2026-08-31
AUv2: probe what a layout does to every bus, not just the main ones

baconpaul:
The configurable-audio-ports probe asked can_apply_configuration about a grid of main-bus channel counts and recorded the pairs that came back true. That is all an AUChannelInfo can carry, so it was enough to advertise the layouts -- but not enough to let a host configure one. A layout is a property of the whole port list: a plugin that goes mono may take its sidechain with it, and can_apply_configuration answers yes or no without saying what the ports would become.

So ValidFormat had nothing to vet a non-main bus against and pinned it to the count snapshotted at PostConstructor. A host that took the advertised mono layout and set its busses got kAudioUnitErr_FormatNot- Supported on the sidechain -- the plugin appearing to reject a configuration it had just offered -- and, since the wrapper reconfigures the ports inside Initialize either way, the AU and the plugin were left disagreeing about that bus's width. Out of process, which is how a sandboxSafe unit is loaded, the v3 bridge cached the stale count and carried on with it. auval does not see any of this: it sets the main busses only, and hosts the unit in process.

The probe now applies each candidate and records every bus, so a layout is what the plugin does rather than what was asked of it, and ValidFormat answers per bus. The plugin is put back into the layout it started in; if it will not go back, the port caches are rebuilt from whatever the ports now are, since a cache that lied about the live layout would make every format the host set look like it already matched and nothing would ever be pushed. That rebuild is the rescan applyConfigurationFromBusFormats was already doing, now shared.

Only plugins implementing clap.configurable-audio-ports probe at all, and the applies are bookkeeping on a deactivated plugin.

Assisted-by: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/src/wrapasauv2.cpp
+++ b/src/wrapasauv2.cpp
@@ -1151,6 +1151,54 @@ std::vector<clap_audio_port_configuration_request_t> WrapAsAUV2::mainBusConfigur
   return requests;
 }
 
+void WrapAsAUV2::recordBusChannelCounts(ChannelCapsCache &caps) const
+{
+  auto ap = _plugin->_ext._audioports;
+  auto pl = _plugin->_plugin;
+  caps.inputBusChannels.clear();
+  caps.outputBusChannels.clear();
+  for (bool isInput : {true, false})
+  {
+    const auto count = ap->count(pl, isInput);
+    for (uint32_t i = 0; i < count; ++i)
+    {
+      clap_audio_port_info inf;
+      ap->get(pl, i, isInput, &inf);
+      (isInput ? caps.inputBusChannels : caps.outputBusChannels).push_back(inf.channel_count);
+    }
+  }
+}
+
+// Ports past the AU element counts (frozen at PostConstructor) are cached but
+// get no bus; the process adapter clamps to the element counts the same way.
+void WrapAsAUV2::refreshPortCachesAndBusses()
+{
+  auto ap = _plugin->_ext._audioports;
+  auto pl = _plugin->_plugin;
+  const size_t numInputElements = Inputs().GetNumberOfElements();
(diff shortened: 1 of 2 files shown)


baconpaul:
OK this is required to get spectrumworx with 'mono + mono side chain -> mono' or 'stereo + stereo side chain -> stereo' working. Without it the side chain was pinned stereo in the mono case.

I need to review this carefully and would appreciate you doing so also. I just got it working and sent it here so you and i can look. Without a careful review its not ready for merge, but it does allow me to do the mono + mono -> mono config.

src/wrapasauv2.cpp
```diff
@@ -1151,6 +1151,54 @@ std::vector<clap_audio_port_configuration_request_t> WrapAsAUV2::mainBusConfigur
   return requests;
 }
 
+void WrapAsAUV2::recordBusChannelCounts(ChannelCapsCache &caps) const
+{
+  auto ap = _plugin->_ext._audioports;
+  auto pl = _plugin->_plugin;
+  caps.inputBusChannels.clear();
+  caps.outputBusChannels.clear();
+  for (bool isInput : {true, false})
```
defiantnerd, line 1160: never seen that before - nifty

defiantnerd (approved):
everything looks sane for me. I would put it to next and maybe test it again in a real context before making it a new main.

src/wrapasauv2.cpp
baconpaul: Oh yeah I have that idiom all over my code. Like `for (float zoomLevel : { 75, 90, 100 ... } )` its fun.

baconpaul:
yeah no need for this on main for a while. lets park it on next and try our plugins. Thanks for the review!

PR #535, opened 2026-09-01
Auv2 legacy migration pr

baconpaul:


baconpaul:
@defiantnerd I kept this as 3 separate commits (although I need all 3) so they can be individually reviewed. I suggest we rebase merge them onto next if you are OK with them.

defiantnerd (approved):
sounds sane for me, too.

defiantnerd (approved):
so this complements the other PR, instead of the id there is an additional mapping. Okay.

defiantnerd:
They look all good for me. It is a specific application, but yes.

PR #536, opened 2026-09-08
AUv2 stability and thread safety

CrushedPixel:
This PR addresses 4 issues I discovered while working on a plugin that serves as an offline audio generator.
If you would like me to split this into several PRs, please let me know.

These commits were made with the help of GPT-6 Astra and manually vetted and adjusted.

`3780585`
First of all, my plugin was not outputting any audio in Logic despite producing samples. 
As I mentioned, my plugin generates audio offline and can play it back on user demand, even when there is no incoming audio. The fix was to clear the "silent" flag from the output when it's set for the input.

`121fe78`
My plugin was erroring during the CLAP activate callback and returned `false`, but clap-wrapper reported it as fine to Logic. This commit properly respects the return value of `activate` and `start_processing`.

`0aedc9c`
Logic activates my plugin on a worker thread, while clap-wrapper can call on_main_thread or access the editor on the main thread. These calls could overlap and access the same plugin state at the same time. `AlwaysMainThread()` only changes the result returned by the thread check, it doesn't prevent this overlap. This commit adds a shared lock so these calls happen one at a time. If the plugin is busy, the idle callback is deferred to the next tick. 

`a1ac38b`
During cleanup, clap-wrapper called gui.destroy twice: once through the editor’s cleanup callback and then again directly.

### DIFF
--- a/src/detail/auv2/auv2_shared.h
+++ b/src/detail/auv2/auv2_shared.h
@@ -29,7 +29,7 @@ namespace free_audio::auv2_wrapper
 typedef struct ui_connection
 {
   uint32_t identifier = kAudioUnitProperty_ClapWrapper_UIConnection_id;
-  Clap::Plugin *_plugin = nullptr;   // points to the plugin instance
+  Clap::Plugin *_plugin = nullptr;  // points to the plugin instance
   /// Serializes host lifecycle calls with editor calls and outlives either connection endpoint.
   std::shared_ptr<ausdk::AUMutex> _mainThreadMutex;
   clap_window_t *_window = nullptr;  // points to a window handle, actually ptr to wrapping NSView class
--- a/src/detail/auv2/auv2_shared.h
+++ b/src/detail/auv2/auv2_shared.h
@@ -14,6 +14,8 @@
 
 #include <iostream>
 #include <functional>
+#include <memory>
+#include <AudioUnitSDK/AUUtility.h>
 #include "clap_proxy.h"
 #include <AudioToolbox/AudioUnitProperties.h>
 
@@ -28,6 +30,8 @@ typedef struct ui_connection
 {
   uint32_t identifier = kAudioUnitProperty_ClapWrapper_UIConnection_id;
   Clap::Plugin *_plugin = nullptr;   // points to the plugin instance
+  /// Serializes host lifecycle calls with editor calls and outlives either connection endpoint.
+  std::shared_ptr<ausdk::AUMutex> _mainThreadMutex;
   clap_window_t *_window = nullptr;  // points to a window handle, actually ptr to wrapping NSView class
   uint32_t *_canary = nullptr;       // a canary in the Windows class
   std::function<void(clap_window_t *, uint32_t *)> _registerWindow = nullptr;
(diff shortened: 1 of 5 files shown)


src/wrapasauv2.cpp
```diff
@@ -173,7 +178,6 @@ WrapAsAUV2::WrapAsAUV2(const std::string &clapname, const std::string &clapid, i
       if (_plugin)
       {
         _plugin->initialize();
-        _os_attached.on();
```
defiantnerd, line 178: why this? attaching to the list of active instances should be maintained

defiantnerd: ah, got it.

src/detail/auv2/auv2_shared.h
```diff
@@ -28,6 +30,8 @@ typedef struct ui_connection
 {
   uint32_t identifier = kAudioUnitProperty_ClapWrapper_UIConnection_id;
   Clap::Plugin *_plugin = nullptr;   // points to the plugin instance
+  /// Serializes host lifecycle calls with editor calls and outlives either connection endpoint.
+  std::shared_ptr<ausdk::AUMutex> _mainThreadMutex;
```
defiantnerd, line 34: I think I would prefer a recursive mutex here, so all "main thread" tasks could be recursive for certain API calls

defiantnerd: AUMutex wraps recursive_mutex.

defiantnerd:
Thanks for the contribution - please run clang-format on your patches, so it can be merged.

CrushedPixel:
Done!

defiantnerd:
there are conflicts :(

baconpaul:
ahh and we don't have write access to resolve them to your branch. Let us know if you can or if we need to.

src/detail/auv2/process.cpp
```diff
@@ -353,6 +355,9 @@ void ProcessAdapter::process(ProcessData &data)
 
   _plugin->process(_plugin, &_processData);
 
+  // A CLAP plugin may generate audible output even when its inputs are silent.
+  data.flags &= ~kAudioUnitRenderAction_OutputIsSilence;
```
defiantnerd: Drop the dead data.flags &= ~OutputIsSilence at process.cpp:359
Render owns that flag now, so it misleads the next reader into thinking it does something.

src/wrapasauv2.cpp
```diff
@@ -246,9 +253,8 @@ OSStatus WrapAsAUV2::Initialize()
   auto guarantee_mainthread = _plugin->AlwaysMainThread();
   if (!activateCLAP())
   {
-    // The plugin refused the host-chosen main-bus format pair, or refused to
-    // activate at all (see activateCLAP). Refuse the initialization rather
-    // than render with buffers sized differently from the plugin's ports.
+    // A rejected configuration or lifecycle call leaves the CLAP deactivated.
+    // Report the failure so the host does not render an unavailable processor.
     return kAudioUnitErr_FormatNotSupported;
```
defiantnerd: `kAudioUnitErr_FailedInitialization` instead of `kAudioUnitErr_FormatNotSupported` for an activation failure, so a host doesn't retry other formats pointlessly.

defiantnerd:
Thanks!

PR #539, opened 2026-09-10
AUv2: offer a plugin the layout where its non-main busses move too

baconpaul:
## The problem

The `PostConstructor` probe sweeps main bus counts and, for every other port,
fills in the count that port currently has. Nothing in AUv2 tells it otherwise —
`AUChannelInfo` describes main busses only, so the probe has no opinion about a
side chain and can only express that as "the same as now". The VST3 wrapper never
faces this: `setBusArrangements` hands it an arrangement for every bus and it
relays them.

But *"I have no opinion about this port"* and *"I require this port to be stereo"*
are the same bytes on the wire, and a plugin can only read the second.

So a plugin whose side chain has to match its main bus — a spectral effect reading
the two channel for channel, say — is asked to go from stereo to mono as
**1 / 2 / 1**, a layout it does not have. It refuses, correctly, and the probe
reads that as "no mono support" and gives up. The plugin advertises stereo alone
in every AU host while working fine everywhere else.

Asking more politely doesn't help either: the extension says a port the requests
do not name keeps what it has, so omitting the side chain leaves it exactly where
an explicit request would. The only way to propose narrowing it is to ask.

## The change

**The probe sends two shapes per candidate** — non-main ports left where they are,
and non-main ports taken along to the main input width — and records whichever the
plugin accepts. A plugin whose side chain is fixed accepts the first, one whose
side chain must follow accepts the second, and one that can do both gets both
recorded. `ValidFormat` admits a bus width that *any* recorded layout gives that
bus, so offering only one of the two is what turns a reasonable host request into
`kAudioUnitErr_FormatNotSupported`.

Recording both means the same main pair can appear twice, which `AUChannelInfo`
has no way to say and no need to, so `SupportedNumChannels` now dedupes.

**`applyConfigurationFromBusFormats` asks for every bus the host set** rather than
the main ones alone. `ValidFormat` vets one scope at a time, because that is how a
host sets formats, so it can only ask whether some recorded layout gives that bus
that width — a host can therefore land on a combination no single layout has.
Asking the plugin for exactly what the host set turns that into a refusal at
activation rather than into an AU element whose channel count the active plugin's
port does not share, which the render path would then write past.

## Verification

Against a spectral effect with a main pair and a side chain that must match it,
driven by a small host that sets bus formats one `SetProperty` at a time the way a
DAW does:

| layout | before | after |
|---|---|---|
| `1 x 1 -> 1` | `kAudioUnitErr_FormatNotSupported` | initialises |
| `1 x 2 -> 1` | — | refused, correctly |
| `2 x 1 -> 2` | — | refused, correctly |
| `2 x 2 -> 2` | initialises | initialises |

`auval` passes and reports `[2, 2] [1, 1]` both before and after — it sets the
main busses and leaves the side chain alone, so it never sees any of this. Also
confirmed working in Logic Pro on a mono track, which was the original report.

### DIFF
--- a/src/detail/auv2/auv2_base_classes.h
+++ b/src/detail/auv2/auv2_base_classes.h
@@ -800,10 +800,11 @@ class WrapAsAUV2 : public ausdk::AUBase,
   void addOutputBus(int bus, const clap_audio_port_info_t *info);
 
   // Configuration requests for the main busses: main ports get the given
-  // channel counts, non-main ports keep their current ones. Shared by the
+  // channel counts, non-main ports keep their current ones -- or, when
+  // nonMainChannels is non-zero, move to that instead. Shared by the
   // PostConstructor probe and applyConfigurationFromBusFormats.
   std::vector<clap_audio_port_configuration_request_t> mainBusConfigurationRequests(
-      uint32_t mainInChannels, uint32_t mainOutChannels) const;
+      uint32_t mainInChannels, uint32_t mainOutChannels, uint32_t nonMainChannels = 0) const;
 
   // If the host chose main-bus stream formats that differ from the current
   // CLAP port layout (ValidFormat admits every probed layout), push the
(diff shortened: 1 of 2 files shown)


baconpaul:
@defiantnerd we need to merge this one to next also. Also side chains with multiple ports are a pain!

I propose we merge this and your preset one both tomorrow. WDYT?

defiantnerd:
that was a trick one! I'll give it a quyick look today and merge them both to next

PR #540, opened 2026-09-11
AUv3: stop asking the host to release a parameter name it never owned

defiantnerd:
CFNameRelease hands the host ownership of the CFStringRef, and the AUv3 to v2 bridge fills that field from the AUParameter's displayName without a matching retain,
so an honouring host over-releases a string it never owned.


baconpaul:
@defiantnerd think we should merge this yeah?

Issue #551, opened 2026-09-14
AUv2: a host parameter set while the CLAP is deactivated is silently dropped


defiantnerd:
Finding AUV2-2 from the 0.16 release review.

## The problem

`src/wrapasauv2.cpp:775`. When the CLAP is deactivated, `_processAdapter` is null, so a host `SetParameter` queues nowhere. `_requestedFlush` is still set, and `flushParameters()` then flushes an empty `_flushAdapter`. Nothing re-syncs AU element values into the plugin when it is next activated, so the value is lost outright.

The comment in `flush()` claims the deactivated path carries exactly this traffic. It does not.

## Failure

A host sets `kAudioUnitProperty_BypassEffect` (which routes through `SetBypassEffect` → `SetParameter`) on the uninitialized unit at project load. After `Initialize`, the plugin is not bypassed while the host UI shows that it is.

## Suggested fix

Queue onto `_flushAdapter` when `_processAdapter` is absent, so the deactivated path actually carries what its comment says it carries.

Issue #538, opened 2026-09-09
AUv3: parameter changes stop syncing in both directions when audio processing pauses


singularpoint:
With a CLAP plugin wrapped as AUv3, parameter changes stop synchronizing in both directions while audio processing is suspended, even though parameters remain editable. Resuming audio processing delivers the pending changes.

Reproduced with AUv3 in **both hosts**:

- **REAPER 7.79**
- **Audio Hijack 4.6.0**

Tested on macOS 26.6.2 / Apple Silicon, using `next` at [`9501dad0`](https://github.com/free-audio/clap-wrapper/commit/9501dad0f5f6449ee007aac67c68048350f1205a), matching upstream `next` when checked.

## Expected behavior

Parameter exchange should continue without requiring audio processing to resume. CLAP explicitly provides `params.flush()` for this purpose. Its parameter contract says host-control changes are delivered through `process()` or `flush()`, and `host_params.request_flush()` requires the host to schedule one of those calls. Here, the wrapper is the CLAP host. [CLAP parameter contract, at the SDK revision used](https://github.com/free-audio/clap/blob/195b42a004144fab0b3cf95e9c067187d15365b7/include/clap/ext/params.h).

`flush()` must not run concurrently with `process()` and must follow CLAP's active/inactive thread rules. Those constraints require serialized delivery; they do not make audio processing a prerequisite for parameter exchange.

## Quick reproduction — no debugger

Attached: **passive-parameters-repro.zip**. It contains a small CLAP plugin, AUv3 build/install instructions, and detailed reproduction steps. The plugin has one parameter, silent audio output, exact-value buttons, and an event log.

The following steps use Audio Hijack's block **off/on** control as a concrete example. Build/install the plugin, load **Example: Passive Parameters**, and start the session with the block **on**. Locate both the plugin editor and the host's own **Value** control.

1. Click **Set 0.25** in the plugin editor; confirm both sides show 0.25.
2. Turn the block **off**, click **Set 0.75**, and wait five seconds.
   **Plugin: 0.75; host: still 0.25.**
3. Turn the block **on**, without another edit. The host catches up to 0.75.
4. Reset both sides to 0.25 while on, then turn the block **off** again.
5. Change Value through the **host's** control to approximately 0.6 and wait.
   **Host: new value; plugin: still 0.25.**
6. Turn the block **on**. The plugin catches up.

[...]

Issue #550, opened 2026-09-14
AUv2: three edge cases in Render() worth a second look


defiantnerd:
Three related observations in `WrapAsAUV2::Render()` (`src/wrapasauv2.cpp`), collected while fixing AUV2-1 in the 0.16 release review. Grouped because they all concern the same function.

## 1. A host passing action flags gets silence

`Render()` gates on `if (_initialized && (inFlags == 0))`. AUV2-1 added an `else` branch that zeroes the output and raises `kAudioUnitRenderAction_OutputIsSilence` — which fixed the stale-audio bug, but it also means a host passing e.g. `kAudioUnitRenderAction_DoNotCheckRenderArgs` is now silently *muted* rather than silently *stale*.

Arguably still wrong: those flags do not mean "do not process". There is an `assert(inFlags == 0)` at the top of the function recording the expectation that no host does this, but an assert is not a plan for release builds.

**Suggested fix:** decide which flags are genuinely incompatible with processing and handle the rest normally, rather than treating any non-zero `inFlags` as unrenderable.

## 2. `OutputIsSilence` can leak in from upstream

`ProcessData::flags` is a reference passed through to `PullInput`. An upstream unit setting `kAudioUnitRenderAction_OutputIsSilence` on its own output can therefore surface as *this* wrapper's output flag, telling the host the plugin produced silence when it did not — a synth fed by a silent input bus, for instance.

**Suggested fix:** take a local copy of the flags for the input pull and set the wrapper's own output flag deliberately.

## 3. The render thread spins on a lock the main thread can hold indefinitely

(This is AUV2-3 from the review.) `Render()` spins on `_processLock`, and `onIdle()` can hold that lock across `clap_plugin_params.flush()` — plugin code of unbounded duration. If the main thread is preempted while holding it, the render thread spins a whole scheduler quantum.

Scenario: Logic idles a track for more than three blocks, `onIdle()` starts a flush, the user presses play at that moment.

**Suggested fix:** `try_lock` in `Render()` and skip-and-zero on failure, rather than a blocking spin. AUV2-1 already added a correct "produce silence" path this can reuse.

## Notes

macOS-only; none of this was compiled or run here.

Part of the 0.16 review: https://claude.ai/code/artifact/2852de97-bdd0-4364-8c89-6892776b03d5

PR #556, opened 2026-09-14
Keep parameters in sync while the host is not rendering (AUv3 #538, AUv2 #551)

defiantnerd:
Closes #538. Closes #551.

The same bug in two formats: parameter traffic is gated on a processing path that is not running, when CLAP provides `flush()` precisely so that it need not be.

## AUv3 (#538)

Parameter changes stopped moving **in both directions** whenever a host stopped rendering without deallocating — Audio Hijack with its block off, REAPER on an idle track — and resumed only when processing did.

- **Outbound:** `param_request_flush()` only logged, so a value the plugin's editor changed sat in its queue indefinitely. It now records the request and the main-queue idle timer services it.
- **Inbound:** the observer and bypass setter decided *delivery* from `renderResourcesAllocated`. Apple documents that as resource-allocation state, and a host may hold resources for a unit it is not rendering — so it cannot answer whether a render will come. It now selects only the *path* (adapter queue vs. direct flush), and the queue is drained by the idle tick when no render drains it first.

What decides is the absence of renders, not a flag: the render block records that it ran, and the idle timer flushes once three block periods have passed without one, sized from the sample rate and `maximumFramesToRender`. A rendering host never reaches the flush path.

Adds `Clap::AUv3::ProcessAdapter::flush()`, which drains the host queue into input events, calls the plugin, and forwards parameter and gesture output events to the automation sink. Note and MIDI output are dropped there deliberately — the AU MIDI output blocks may only be called from inside a render.

## AUv2 (#551)

`SetParameter` queued onto the process adapter, which does not exist while the CLAP is deactivated, so the value was dropped outright. A host restoring a project sets `kAudioUnitProperty_BypassEffect` before it calls `Initialize`, so the plugin came up unbypassed while the host UI showed it bypassed.

Delivered two ways, because one is not enough: the idle tick handles it where there is time, and `activateCLAP()` handles it before `clap_plugin.activate()`, because Logic sets bypass and calls `Initialize` back-to-back with no idle tick in between.

## Threading

`clap_plugin_params.flush()` must not run concurrently with `process()`, and is `[active ? audio-thread : main-thread]`. AUv3 gains a spin lock the way AUv2 already has one: the render block holds it around `process()` and every wrapper-issued flush holds it too. The activate/deactivate pairs are inside the same lock, so `_initialized` read under it is exactly "the plugin is active" and the thread identity cannot come from stale state — which also closes a pre-existing window where a render could reach `process()` between the adapter being published and `start_processing()`.

The render block takes the lock *after* announcing itself in `_renderInFlight`, and deallocation drains that counter *without* holding the lock, so the two cannot form a cycle.

**Known trade-off, shared with AUv2:** the render thread can spin on a lock the main thread holds across plugin code. The three-blocks-without-a-render heuristic means a rendering host does not reach it, but it is the same pattern #550 raises against AUv2. If a better one is found, both formats should move together.

## Verification

The reporter's six steps in #538 come with a repro plugin. Specifically:

- **Step 2** (plugin edit, block off): the host should catch up within roughly three block periods, *without* step 3.
- **Step 5** (host edit, block off): the plugin should reflect the new value on the same timescale, *without* step 6.
- Normal rendering unaffected — automation timing, no glitches at block on/off transitions.
- Bypass toggled with the block off reaches the plugin.

For AUv2: bypass an effect in Logic, save, reopen — it must start bypassed with the header button agreeing. Plus `auval -v aufx`, and automating a parameter on a muted track before unmuting.

Needs a macOS build and a run through the above before merging.

## Not addressed

The mirror image on the AUv2 side: events queued on the process adapter between the last render and `Cleanup` or a restart still die with the adapter. AUv3 drains those after `deactivate()`; AUv2 could do the same. Separate change.


singularpoint:
Thanks for picking this up! A few potential follow-ups from my investigation: respecting CLAP_PARAM_REQUIRES_PROCESS in active flushes, ensuring inactive flushes run on the actual main thread, and handling output queue overflow. For the spinlock trade-off already noted, os_unfair_lock could help mitigate priority inversion on resume.
These may fit better as separate changes to keep this fix focused.

singularpoint:
I confirm the issue #538 is fixed by this PR both in MacOS and iOS.

defiantnerd:
Thanks for bringing it up and that you already checked `next` before creating the issue. Also, thanks for the Repro.

baconpaul (commented):
I'm less familiar with the auv3 code but this seems to match gthe auv2 code in spirit so I'm fine for it to be merged if you're convinced it is correct @defiantnerd

Issue #525, opened 2026-08-21
os::getTickInMS() returns CPU time, not wall-clock time, on macOS and Linux — CLAP timers fire at the wrong rate


defiantnerd:
### Summary

`os::getTickInMS()` returns **process CPU time**, not wall-clock time, on macOS and Linux. Every
CLAP timer registered through the VST3, AAX and AUv3 wrappers is therefore scheduled against the
wrong clock.

```cpp
// src/detail/os/macos.mm:118
uint64_t getTickInMS()
{
  return (::clock() * 1000) / CLOCKS_PER_SEC;   // CPU milliseconds
}

// src/detail/os/linux.cpp:193
uint64_t getTickInMS()
{
  return clock();                                // CPU time in CLOCKS_PER_SEC units (µs on glibc)
}

// src/detail/os/windows.cpp:179
uint64_t getTickInMS()
{
  return GetTickCount64();                       // correct: wall clock
}
```

`clock()` measures CPU time consumed by the process, summed over all threads. So the tick

* **crawls while the process is idle** — timers fire far later than their period, or never;
* **can run faster than real time while the audio threads are busy** — with several busy threads,
  `clock()` advances at a multiple of wall-clock, so timers fire too often;
* is off by a factor of 1000 on Linux/glibc on top of that (`CLOCKS_PER_SEC` is 1000000 there, and
  the value is returned unscaled).

`clap/ext/timer-support.h` specifies `period_ms` in real-time milliseconds, so all three call sites
are wrong through no fault of their own.

### How I ran into it

A CLAP plugin that registers a 4000 ms timer in an out-of-process AUv3 appex never received
`on_timer` at all. Instrumenting the wrapper showed the machinery was healthy — `fireTimers()` ran
every 10 ms for 13.5 s with one live slot and a valid `_ext._timer` — but `to.nexttick <= now` never
became true, because an idle appex burns only a few ms of CPU in that time.

### Affected call sites

All of them compare a `nexttick` computed from `getTickInMS()`:

| | |
|---|---|
| VST3 | `src/wrapasvst3.cpp:1403`, `:1415`, `:1535` |
| AAX | `src/detail/aax/wrapper.cpp:662`, `:1116` |
| AUv3 | `src/detail/auv3/auv3_audiounit.mm:546`, `:554`, `:584` |

AUv2 registers timers into the same kind of slot list but services them from its own idle pump, so
it is affected only where it shares this helper.

Windows is unaffected.

### Fix

[...]

Issue #523, opened 2026-08-21
AUv2: mark_dirty() is a no-op


baconpaul:
### Summary

`WrapAsAUV2::mark_dirty()` is an empty function body, so a hosted CLAP's call to `clap_host_state::mark_dirty` is dropped under AUv2.

Line references are against `be1a04c` (current `main`).

```cpp
// src/detail/auv2/auv2_base_classes.h:581
void mark_dirty() override
{
}
```

### Why it matters

This is the mildest of the three AUv2 `IHost` stubs I ran into — in practice Logic persists AU state on project save regardless of any notification, so the usual "my session won't be saved" failure doesn't happen. What is lost is the narrower signal: the plugin telling the host that its state no longer matches what the host last read, so the host can update or clear the preset name it is displaying and know its cached `kAudioUnitProperty_ClassInfo` is stale.

### Current state per format

| format | `mark_dirty()` | |
|---|---|---|
| VST3 | `src/wrapasvst3.cpp:1342` → `componentHandler2->setDirty(true)` | ✅ |
| AAX | `src/detail/aax/wrapper.cpp:1101` → `++mNumPlugInChanges` | ✅ |
| **AUv2** | `src/detail/auv2/auv2_base_classes.h:581` — empty | ❌ |
| AUv3 | `src/detail/auv3/auv3_audiounit.mm:184` — logs only | ❌ |
| standalone | `src/detail/standalone/standalone_host.h:109` | n/a |

### Proposed fix

AU has no exact equivalent of `setDirty`, but the wrapper already uses the nearest idiom elsewhere: `WrapAsAUV2::param_rescan()` issues `PropertyChanged(kAudioUnitProperty_ClassInfo, kAudioUnitScope_Global, 0)` (`src/wrapasauv2.cpp:492`) to say that the saved state has moved.

So `mark_dirty()` could reasonably be:

```cpp
void mark_dirty() override
{
  PropertyChanged(kAudioUnitProperty_ClassInfo, kAudioUnitScope_Global, 0);
}
```

possibly with `kAudioUnitProperty_PresentPreset` alongside it, if the intent is for the host to stop showing a preset name the state no longer matches.

I'm less sure of the right mapping here than for the other two, so this may equally well be resolved as "AU has no equivalent" plus a comment saying so — an empty body with no explanation is what made me file it, since the neighbouring stubs turned out to be real gaps.

[...]

Issue #522, opened 2026-08-21
AUv2: param_request_flush() is a no-op and ProcessAdapter::flush() is never defined


baconpaul:
### Summary

`WrapAsAUV2::param_request_flush()` is an empty function body, so `clap_host_params::request_flush` does nothing in AUv2. A plugin that queues a parameter change and asks the host to come and collect it never gets collected, unless a render happens to come along.

Relatedly, `Clap::AUv2::ProcessAdapter::flush()` is **declared but never defined** — there is no AUv2 flush path to call even if the request were wired up.

Line references are against `be1a04c` (current `main`).

```cpp
// src/detail/auv2/auv2_base_classes.h:608
void param_request_flush() override
{
}
```

```cpp
// src/detail/auv2/process.h:103
void flush();          // no definition anywhere under src/detail/auv2/
```

### Why it matters

`request_flush` exists for the case where there is no audio thread to carry the change: transport parked, track muted, plugin offline. A clap-first plugin that routes edits to its engine through the process-time event queue — which is the shape the wrapper's own `ProcessAdapter` encourages — has no other way to get them delivered.

The concrete case downstream: SpectrumWorx loads a preset on the UI thread, queues the resulting parameter values, and calls `request_flush` so they arrive even with the transport stopped. Under AUv2 that request is dropped.

### Current state per format

| format | `param_request_flush()` | flush path |
|---|---|---|
| VST3 | `src/wrapasvst3.cpp:1290` sets `_requestedFlush`, consumed at `:1483` | builds a `Clap::ProcessAdapter` and calls `pa.flush()` |
| AAX | `src/detail/aax/wrapper.cpp:1050` sets `_flushRequested`, consumed at `:1204` | `AAXProcessAdapter::flush()`, `src/detail/aax/process.cpp:418` |
| **AUv2** | `src/detail/auv2/auv2_base_classes.h:608` — empty | declared at `src/detail/auv2/process.h:103`, never defined |
| AUv3 | `src/detail/auv3/auv3_audiounit.mm:442` — logs only | — |
| standalone | `src/detail/standalone/standalone_host.h:154` — empty | — |

### Proposed fix

The AAX wrapper is the closest model and is already the two-piece version of this:

1. `param_request_flush()` sets an `std::atomic_bool _flushRequested`.
[...]

Issue #390, opened 2025-05-01
If now proper audio device exists, wrapper skips activate() but calls deactivate()


defiantnerd:
When using the standalone wrapper and there is no proper audio device present to be opened via WASAPI, an exception is thrown and the plugin does not get an activate() call.

But on exiting the program, deactivate is being called - which triggers clap-helpers framework detecting imbalanced activate/deactive calls.

defiantnerd:
fixed with #506

PR #559, opened 2026-09-14
README: document the contribution flow and the WebAssembly target

defiantnerd:
Two gaps in the README, both of which cost a newcomer (or an agent) a wrong assumption.

**Contribution.** Nothing said that development happens on `next` and that `main` only moves at a release. The more expensive omission: issues here are not closed when a fix lands, so an open issue is not evidence of an unfixed bug — #390, #521, #522, #523 and #525 are all fixed on `next` and still open. Someone checking GitHub before reporting would conclude the opposite.

**WebAssembly.** The WASM path was undocumented outside the CMake. It is easy to assume it means "a VST3 in the browser", when in fact nothing is wrapped: the build compiles the CLAP itself into a `.wclap`, and VST3, AUv2, AUv3, AAX and standalone are all unavailable there.

Everything in the WASM paragraph comes from the build files:

- `cmake/make_clapfirst.cmake:74-91` — `PLUGIN_FORMATS` defaults to `WCLAP` alone, other formats forced off
- `cmake/shared_prologue.cmake:56-61` — Emscripten / `wasm32` / `wasm64` detection
- `cmake/make_clapfirst.cmake:344-352` — `-msimd128`, `clap_entry` and `malloc` exports
- `cmake/wrap_wclap.cmake:25-51` — bundle vs. flat output layout

Documentation only; no code or build changes.

One caveat worth a maintainer's eye, deliberately left out of the README: `shared_prologue.cmake:61` sets `ANY_WASM_TOOLCHAIN` with `PARENT_SCOPE`, while line 64 tests it in the current scope, where that leaves it unset. Reproducing the same scope structure in a throwaway CMake project, CMake warns and the `EMS` branch is not taken. If that is a real bug, this README now describes intended rather than actual behaviour — happy to open a separate issue.


baconpaul:
happy to merge this

baconpaul (approved):

PR #560, opened 2026-09-14
Tell a wrapped plugin which format it is in, and which host

baconpaul:
A CLAP inside clap-wrapper can't currently tell that it is wrapped, or as what. `host->name` only half answers who the host is: it carries a ` (CLAP-as-VST3)` style suffix, the Apple formats add a version, and an AUv3 gets the name of its own appex.

This adds a host extension, `clap-wrapper.host-information/0` in `include/clapwrapper/wrapper_host.h`, which every wrapper now provides:

```cpp
auto *cwh = (const clap_wrapper_host_information_t *)host->get_extension(
    host, CLAP_WRAPPER_HOST_INFORMATION);
if (cwh)
{
  if (strcmp(cwh->get_wrapper_flavor(host), CLAP_WRAPPER_HOST_FLAVOR_AUV2) == 0)
    ...
  const char *name = cwh->get_underlying_host_name(host); // "REAPER", or null
}
```

A null extension means a real CLAP host.

### `get_wrapper_flavor`

One of `CLAP_WRAPPER_HOST_FLAVOR_AUV2`, `_AUV3`, `_AAX`, `_VST3`, `_STANDALONE`. `Clap::IHost::wrapper_flavor()` is pure virtual, so a new wrapper has to answer. The AAX scan-time microhost answers `aax` too.

### `get_underlying_host_name`

The host's own name without the wrapper suffix, or null when it can't be known.

| Wrapper | Answer |
|---|---|
| VST3 | `IHostApplication::getName` |
| AAX | `AAX_IController::GetHostName`; null in the scan-time microhost |
| AUv2 / AUv3 | the name of the process the plugin was loaded into; null when that is an XPC service (`AUHostingService` for an out of process AUv2, the appex for an AUv3) |
| Standalone | `Wrapper Mac Standalone` / `Wrapper Windows Standalone` / `Wrapper Linux Standalone` |

Each wrapper stores the name before creating the plugin and never changes it, so the call is thread-safe.

`host->name` is deliberately unchanged. REAPER as a native CLAP host also calls itself `REAPER`, and a plugin special-casing that would start doing so inside the VST3. VST3 and AAX now build `host->name` from the stored name, so the two can't disagree.

### Example

The clap-first distortion example logs its flavor (comparing against a few of the constants) and the underlying host name.

### Testing

- Builds on macOS: VST3, AUv2, standalone and AAX with Ninja, AUv3 with Xcode. clang-format clean.
- VST3 validator: 47/47, and the example logs `Wrapped as a VST3` / `Underlying host is 'vstvalidator'`.
- Not run: AUv2, AUv3, AAX or standalone at runtime, including the XPC check. Not built on Windows or Linux.


baconpaul:
@defiantnerd as discussed. A way to tell if you are wrapped by a host query.

defiantnerd (approved):

